PR #38, opened 2021-10-15
cancelled streamsubscriptions on client close

kwekel:
This change causes the streamsubscriptions to be closed when the function close is called on the CoapClient. This new code allows creating a new client with a new host if the previous client is closed. Note that this means that having multiple clients at the same time is still not possible.


shamblett:
Thanks, OK with this, package re released at version 3.1.2

PR #43, opened 2021-10-28
fix: Token counter not increasing between requests

Sorunome:
Fixes the token counter not increasing between requests, thus breaking concurrent requests


shamblett:
Thanks, package re published at version 3.2.1.

BTW please feel free to create an AUTHORS file in the top level and add your details if you wish in one of your next updates.

PR #46, opened 2021-12-05
feat: update Content Format registry

JKRhb:
This PR updates the Content Format registry in `coap_media_type.dart` by removing a number of outdated entries and adding a number of additional ones (cf. the respective [IANA registry](https://www.iana.org/assignments/core-parameters/core-parameters.xhtml#content-formats)). It also updates the static `fileExtension` method so that it can deal with cases where no file extension is defined.

There are a couple of minor issues that might have to be solved before merging the PR. One concerns the question of how to deal Content Formats where parameters are essential (e. g. `application/cose; cose-type="cose-encrypt"` and `application/cose; cose-type="cose-mac"`). Those haven't been added to the registry yet. Another one concerns the file extensions and the value that is returned if none is defined.

I am looking forward to your review!


JKRhb:
I removed `coap_draft03.dart` for now as it prevented building the project. Let me know if you want to keep it, then we might need to find a workaround.

shamblett:
OK, looks good so far, removal of draft 3 is fine, I doubt if anyone uses it or ever will. 

The minor details/TODO's that are there don't need to be resolved now, if you have a plan for this fine, if not we can leave the cose stuff out for now and fix it later, same for the undefined file extension, just pick a suitable one, we can always document what we have done somewhere so people know, or even let them specify their own in the COAP config although this may be too much work for what its worth.

Its always a balance between being as correct as possible or releasing a new version of the package that will be of more use to users and fixing more minor details later.

JKRhb:
Thank you for the quick response!

> OK, looks good so far, removal of draft 3 is fine, I doubt if anyone uses it or ever will.
> 
> The minor details/TODO's that are there don't need to be resolved now

Okay, great :) I am going to leave some minor comments regarding things I noticed since I opened the PR once these are resolved we should be able to merge I guess.

> Its always a balance between being as correct as possible or releasing a new version of the package that will be of more use to users and fixing minor details later.

That's true! I will probably keep working with your package so I will hopefully be able to provide a follow-up to this PR later :)

lib/src/coap_media_type.dart
```diff
@@ -81,54 +88,119 @@ class CoapMediaType {
...
   static const int any = 0xFF;
 
   /// Checks whether the given media type is a type of image.
   /// True iff the media type is a type of image.
   static bool isImage(int mediaType) =>
-      mediaType >= imageGif && mediaType <= imageTiff;
+      mediaType >= imageGif && mediaType <= imagePng;
 
   /// Is the media type printable
   static bool isPrintable(int? mediaType) {
```
JKRhb, line 199: I guess the new JSON and XML based formats could probably be added here, right? (Also `textCss` and `applicationJavascript` I guess?) Are there any other criteria for a media type to be printable?

lib/src/coap_media_type.dart
```diff
@@ -149,7 +221,11 @@ class CoapMediaType {
   /// Gets the file extension of the given media type.
   static String fileExtension(int mediaType) {
     if (_registry.containsKey(mediaType)) {
-      return _registry[mediaType]![1];
+      if (_registry[mediaType]!.length > 1) {
+        return _registry[mediaType]![1];
+      } else {
+        return 'undefined';
```
JKRhb, line 227: I had a another look at the IANA registry but there does not seem to be an explicit "undefined" extension. Would returning `null` be an option for future versions? For now the API change is probably not desirable.

JKRhb (commented):
Here are two small comments that might be worth discussing.

JKRhb:
One last thing that is not related to the changes in this PR:

The `name` method

```dart
  /// Returns a string representation of the media type.
  static String name(int mediaType) {
    if (_registry.containsKey(mediaType)) {
      return _registry[mediaType]![0];
    } else {
      return 'unknown/$mediaType';
    }
  }
```

returns `unknown/$mediaType` at the moment for unknown media types. However, I read during my IANA related research that `application/octet-stream` might be a better choice for data that cannot be identified regarding its content type as it is labelled as a generic bytestream this way. Do you think this could be an improvement? Or would you prefer the current approach?

lib/src/coap_media_type.dart
shamblett: YEp sounds OK, I guess the only criteria is that the output is readable when printed.

lib/src/coap_media_type.dart
shamblett: Now we are in the world of null safety I would prefer not to return a null, I've got into the habit now of returning an empty string if no other value can be calculated. You could just annotate the API comment to say that 'undefined' can be returned if the media type cannot be resolved, as long as we inform people what we are doing I think its OK.

shamblett:
application/octet-stream is OK by me, again add a line in the API comment above the method explaining that this is the default if no other media type can be found, people will then at least know the behaviour.

JKRhb:
Thank you for the feedback! I think this PR should now be mergeable :)

shamblett:
OK, done, package re released at version 3.3.0, the TODO's and fix me's can be done later.

PR #48, opened 2021-12-12
fix: use compressed bytes for block option encoding

### LINKED ISSUE #47: Blockwise transfer fails when block number exceeds 15
When performing a (Block2) blockwise transfer, the block numbering seems to jump to 128 instead of 16 after the block with number 15 has been received. At the same time, (in my case) the block size jumps from 16 to 32 indicating that there seems to be an issue with the binary encoding. I will try to provide a code example for reproducing the issue.

JKRhb:
This PR resolves #47 by using the correct getter for block options during encoding. See #47 for more context.

### DIFF
--- a/lib/src/codec/encoders/coap_message_encoder18.dart
+++ b/lib/src/codec/encoders/coap_message_encoder18.dart
@@ -60,7 +60,11 @@ class CoapMessageEncoder18 extends CoapMessageEncoder {
       }
 
       // Write option value
-      writer.writeBytes(opt.valueBytes);
+      if (opt is CoapBlockOption) {
+        writer.writeBytes(opt.blockValueBytes);
+      } else {
+        writer.writeBytes(opt.valueBytes);
+      }
 
       lastOptionNumber = optNum;
     }


JKRhb:
Note: I have implemented the change only in the Draft 18 (or rather: RFC 7252) encoder for now. 

Following our discussion in #46, I think the implementations of the older drafts could probably be removed in general?

shamblett:
Thanks, I've re published the package at version 3.3.1 for this fix.

Agree that maintaining the old drafts is not something needed moving forward, we just need draft 18, raised #49 to address this.

You could address your comment on #47 by continuing with this issue and raise another PR or close this now and raise another issue to look at options encoding in general, whatever you prefer.

JKRhb:
Thank you for merging this PR and publishing the new version! Unfortunately, the fix did not work out as expected and the issue still persists. Considering https://github.com/shamblett/coap/issues/47#issuecomment-992009156 this seems to be in fact a more general problem with the encoding of larger options. As you suggested, I will raise another issue to deal with this :)

Also, thank you for opening #49!

Issue #49, opened 2021-12-13
Remove unused CoAP Draft Specifications


shamblett:
Going forward only CoAP IETF Draft 18 is going to be supported, the other older drafts should be removed.

The package should be updated to remove the older drafts, remove the configuration option to select the specification and the README updated to state that only Draft 18 is now supported.

shamblett:
Note the comment from #51 -
```
It took me a couple of minutes but I now realized that the behavior of this test changed because the content-format is encoded as a string value and not as a the numeric representation from the registry, hence the mixed up bytes. I think we should open an issue for this or mention it in #49 as changing the encoded value to a numeric one messes up the tests for the older drafts :/ After the old drafts have been removed we could start fixing/updating the tests as a whole.
```
This should be addressed either as part of this issue or another issue raised as suggested.

PR #51, opened 2021-12-22
fix: use correct byte order for numeric options

### LINKED ISSUE #50: Encoding of options is malformed once they reach a length of three bytes
Options that are at least three bytes long are currently not correctly encoded. This is at least the case with the Accept and Block1/Block2 options. 

After doing some more research I now think I discovered the real reason why this is happening: The bytes after the one for option delta and length are using an incorrect byte order which is why they get malformed. ~~This should hopefully be relatively easy to fix.~~ 

Edit: The fix has been a bit more challenging than expected. I now realized that only numeric options like Block1, Block2, and Accept have a wrong byte order both during encoding and decoding while string-based options like uriPath are not affected. I will have another look into this.

(This issue is related to https://github.com/shamblett/coap/issues/47 where I first encountered this problem in the context of blockwise transfer.)

JKRhb:
This PR now finally fixes #50. The reason for the bug comes from [RFC 7252, section 3.2](https://datatracker.ietf.org/doc/html/rfc7252#section-3.2) which specifies that numeric option values have to be encoded in network byte order which has not been the case yet. With this patch, block-wise transfer with more than 16 blocks and content-formats with values greater than 255 now work as expected :)


test/coap_message_encode_decode_test.dart
```diff
@@ -564,16 +564,16 @@ void main() {
...
-          101,
           120,
+          101,
           116,
-          47,
-          112,
-          108,
-          97,
-          105,
-          110,
```
JKRhb, line 576: It took me a couple of minutes but I now realized that the behavior of this test changed because the content-format is encoded as a string value and not as a the numeric representation from the registry, hence the mixed up bytes. I think we should open an issue for this or mention it in #49 as changing the encoded value to a numeric one messes up the tests for the older drafts :/ After the old drafts have been removed we could start fixing/updating the tests as a whole.

shamblett:
Thanks, the network byte ordering is a good catch, these can be difficult to spot, I've been caught with this before. Package re published at version 3.3.2.

I've updated #49 with your comment above so it will not be lost.

PR #52, opened 2022-01-02
Remove unused CoAP Draft Specifications, update tests

### LINKED ISSUE #49: Remove unused CoAP Draft Specifications

JKRhb:
This PR removes the old drafts from the library, renames the Draft 18 specification to RFC 7252, and updates the tests accordingly.

Maybe the naming scheme needs to be adjusted a little – let me know if anything should be changed :)

Closes #49.


shamblett:
Naming looks OK to me, lets see if any users comment on it.

Package re published at version 3.4.0

PR #53, opened 2022-01-02
chore: replace http links with https in README

JKRhb:
This PR replaces two http links in the README file with https ones which should increase the package score on pub.dev by five points to the maximum value of 120 (for native packages, that is).


shamblett:
Thanks, been meaning to do this

PR #56, opened 2022-01-23
fix: switch to unicast for blockwise transfer after multicast request

### LINKED ISSUE #55: Blockwise transfer during multicast not working properly
If a Client sends a multicast request and receives a response using blockwise transfer, it seems as if the client is currently not switching to a unicast transfer to handle the remaining blocks (c. f. [RFC 7959, section 2.8](https://datatracker.ietf.org/doc/html/rfc7959#section-2.8)). This causes problems as subsequent blocks will also be requested from the multicast address which leads to an exponential(?) growth of requests and responses (if multiple servers are part of the same group and provide a similar resource) until a final block has been reached while also corrupting the total payload.

I will try to provide a fix for this problem in the upcoming days.

JKRhb:
This PR fixes #55 by adding a method to the blockwise layer that converts multicast to unicast requests during blockwise transfer. To do so, some networking methods are adjusted to be able to retrieve the source address from responses and to provide a different address than the initial one for requests.

### DIFF
--- a/lib/src/stack/coap_blockwise_layer.dart
+++ b/lib/src/stack/coap_blockwise_layer.dart
@@ -213,6 +213,28 @@ class CoapBlockwiseLayer extends CoapAbstractLayer {
     }
   }
 
+  CoapBlockwiseStatus _copyBlockStatus(
+      CoapBlockwiseStatus? oldBlockwiseStatus, int currentSZX) {
+    final newStatus = CoapBlockwiseStatus.withSize(
+        oldBlockwiseStatus?.contentFormat,
+        oldBlockwiseStatus?.currentNUM,
+        currentSZX);
+    newStatus.blocks = oldBlockwiseStatus!.blocks;
+    return newStatus;
+  }
+
+  CoapExchange _convertMutlicastToUnicastExchange(
+      CoapMulticastExchange exchange, CoapRequest block) {
+    final endpoint = exchange.endpoint;
+    final originalRequest = exchange.request;
+    final newExchange = CoapExchange(block, exchange.origin,
+        namespace: exchange._eventBus.namespace);
+    newExchange.originalMulticastRequest = originalRequest;
+    newExchange.endpoint = endpoint;
+    block.type = CoapMessageType.con;
+    return newExchange;
+  }
+
   @override
   void receiveResponse(
       CoapINextLayer nextLayer, CoapExchange exchange, CoapResponse response) {
@@ -250,7 +272,11 @@ class CoapBlockwiseLayer extends CoapAbstractLayer {
         status.currentNUM = nextNum;
         status.currentSZX = block1.szx;
         final nextBlock = _getNextRequestBlock(exchange.request!, status);
-        nextBlock.token ??= response.token; // reuse same token
+        if (exchange is CoapMulticastExchange) {
+          exchange = _convertMutlicastToUnicastExchange(exchange, nextBlock);
+        } else {
+          nextBlock.token ??= response.token; // reuse same token
+        }
         exchange.currentRequest = nextBlock;
         super.sendRequest(nextLayer, exchange, nextBlock);
         // Do not deliver response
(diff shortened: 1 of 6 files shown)


JKRhb:
Unfortunately, something is still missing in this PR. While the correct IP addresses are being used for message exchanges after the first block was received, the client is only communicating with one of the servers while ignoring other responses.

I haven't been able to find the right point where to fix this yet. Do you have an idea where this could be added? So far I got the impression that a new `CoapExchange` needs to be created for each server from the original multicast Exchange in the matcher.

shamblett:
Yes, I believe you need an exchange for each endpoint, which I believe is a responding server, in your sense,  it keeps track of token values etc. for that endpoint. The harder part may be cleaning these up after the transfer has completedd

JKRhb:
> Yes, I believe you need an exchange for each endpoint, which I believe is a responding server, in your sense, it keeps track of token values etc. for that endpoint. The harder part may be cleaning these up after the transfer has completedd

Thank you for your feedback! I will have another look into this issue over the next upcoming days. I guess the logic for cleaning up will probably have to be added in this method, right? https://github.com/shamblett/coap/blob/c8299af8fbcadd4cd26f7317704cd70db1d8f224/lib/src/net/coap_matcher.dart#L230-L280

shamblett:
Yes, this is what I was thinking, looking at your proposed change however where only addresses are switched we may be OK here, I was more worried if when we switched IP addresses we would create another matcher, if we are not doing that then we only have to ensure we clean up at the end of the block transfer which we do OK anyway.

JKRhb:
It took me pretty long but now I finally got a solution for both the "regular" and the blockwise multicast. I still need to do some cleanup but I'll hopefully be able to push it tomorrow :)

JKRhb:
@shamblett Sorry for all the rebase noise :/ This PR as well as #61 are now finally ready for review, though :) #61 now introduces a working solution for "regular" multicast while this PR extends it to also support blockwise (i. e. block2) transfers.

I tested both PRs thoroughly with two boards that feature both "simple" CoAP endpoints (returning a small payload in response to a GET request or toggeling a lamp after a POST request) as well as larger payloads that are split over multiple exchanges. Let me know if you see any points that could be improvd :)

lib/src/stack/coap_blockwise_layer.dart
```diff
@@ -250,7 +253,12 @@ class CoapBlockwiseLayer extends CoapAbstractLayer {
         status.currentNUM = nextNum;
         status.currentSZX = block1.szx;
         final nextBlock = _getNextRequestBlock(exchange.request!, status);
-        nextBlock.token ??= response.token; // reuse same token
+        if (multicast) {
+          nextBlock.destination = response.source;
+          nextBlock.uriHost = response.source?.address.host;
+        } else {
+          nextBlock.token ??= response.token; // reuse same token
+        }
```
JKRhb, line 261: Hmm, I just noticed that this code is probably not working properly yet (I focussed on block2 transfers for now). Maybe it would be better to remove for now until and add a working solution in a follow-up PR.

JKRhb: I think I now found a better solution by refactoring and reusing the conversion from multicast to unicast :) I haven't tested it with block1 transfers yet, though. Let me know if this solution works for now – in the case of unicast requests everything should still work as before.

shamblett:
OK, I've merge #61 and as far as I can see all is OK, all the examples work OK so do you want this merged now or are you still working on it? I can merge it and test locally if you like.

JKRhb:
> I can merge it and test locally if you like.

That sounds good :) The PR should be ready, the only uncertainty I have is regarding Block1 transfer as I didn't have an opportunity to actually test it yet. Therefore, this could also be excluded from the PR for now.

JKRhb:
(I performed a another rebase to to clean up the commit history a bit.)

PR #60, opened 2022-01-24
chore: replace travis CI with Github Actions

### LINKED ISSUE #57: Move CI to Github Actions?
To make CI results a bit more accessible from the Github UI, it could be moved from Travis to Github Actions. Would it make sense to do so? Or do you prefer testing on Travis? I could provide a PR for the switch if you'd like.

JKRhb:
This would resolve #57 if you want to move the CI to GitHub Actions.


shamblett:
Interesting one this, the reason I use use Travis is that when I published my firdt package(back in 2013) there was of course no  GitHub actions and Travis at the time supported Dart so Ive stuck with it for all my packages since with very little trouble I may add. That in itself of course is not reason enough to never change, one of its problems is visibility, only I can log in and see stuff etc. 
I'll merge this and see what it provides, if it does the same job as Travis I'll keep it and migrate myu othe rpackages to GitHub.

PR #61, opened 2022-02-09
fix: apply fixes to basic multicast scenario

### LINKED ISSUE #55: Blockwise transfer during multicast not working properly

JKRhb:
This PR applies a couple of fixes/changes to the project in order to make it possible to use simple (i. e. non-blockwise) multicast requests. To achieve this, multicast requests are no longer being able to be sent as confirmable messages (c. f. [RFC 7252, section 8.1](https://datatracker.ietf.org/doc/html/rfc7252#section-8.1)). Instead, all confirmable multicasts requests are converted to non-confirmable ones. Furthermore, multicast requests are no longer being labelled as complete after the first response and multiple responses no longer marked as duplicates.

I guess the solution is probably not perfect yet (and does not resolve #55 yet), but I guess it is a step into the right direction. Feedback is highly appreciated :)

### DIFF
--- a/lib/src/net/coap_matcher.dart
+++ b/lib/src/net/coap_matcher.dart
@@ -237,6 +237,20 @@ class CoapMatcher implements CoapIMatcher {
     final keyToken = CoapKeyToken(response.token);
     final exchange = _exchangesByToken[keyToken];
     if (exchange != null) {
+      if (exchange is CoapMulticastExchange) {
+        _log!.info('Matcher - Received response to multicast '
+            'request ${exchange.request}');
+        if (!exchange.alreadyReceived(response)) {
+          _log!.info('Matcher - Received multicast response is from a new '
+              'endpoint.');
+          exchange.responses.add(response);
+        } else {
+          _log!.info('Matcher - Received multicast response is a duplicate.');
+          response.duplicate = true;
+        }
+        return exchange;
+      }
+
       // There is an exchange with the given token
       final prev = _deduplicator!.findPrevious(keyId, exchange);
       if (prev != null) {
--- a/lib/coap.dart
+++ b/lib/coap.dart
@@ -136,6 +136,8 @@ part 'src/stack/coap_chain.dart';
 
 part 'src/stack/coap_layer_stack.dart';
 
+part 'src/net/coap_multicast_exchange.dart';
+
 part 'src/stack/coap_observe_layer.dart';
 
 part 'src/stack/coap_reliability_layer.dart';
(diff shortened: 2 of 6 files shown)


shamblett:
OM I've merged some of your earlier PR's and re released the package at version 3.4.1. I'll look at this in relation to #55 hopeully shortly.

lib/src/net/coap_matcher.dart
```diff
@@ -236,10 +236,14 @@ class CoapMatcher implements CoapIMatcher {
     var keyId = CoapKeyId(response.id);
     final keyToken = CoapKeyToken(response.token);
     final exchange = _exchangesByToken[keyToken];
+    final multicast = exchange?.request?.multicast ?? false;
     if (exchange != null) {
       // There is an exchange with the given token
       final prev = _deduplicator!.findPrevious(keyId, exchange);
       if (prev != null) {
+        if (multicast) {
+          return prev;
```
shamblett, line 245: Maybe a log line here saying we are returning prev because of multicast processing, may help in debugging this at run time.

shamblett:
Nothing strikes me as being odd here but I guess that eventually we will have to go with something and get feedback as to actual behavior by testing, its a difficult area to work in.

It almost looks as though we need a special multicast endpoint maybe, rather than varying the behavior of the general one then of course we would have to instantiate when needed which may be problematic in itself.

Lets see where we go with the changes so far.

lib/src/net/coap_matcher.dart
JKRhb: Good point! Maybe something like this? 

```suggestion
          _log!.info('Matcher - Reuse exchange for multicast processing');
          return prev;
```

JKRhb:
> It almost looks as though we need a special multicast endpoint maybe, rather than varying the behavior of the general one then of course we would have to instantiate when needed which may be problematic in itself.

That is a good idea! This might also make it easier to deal with the blockwise case. I will try it out in the context of #56.

PR #35, opened 2021-10-11
feat: add dtls

Sorunome:
This PR adds DTLS support. It currently relies on an internal, not yet published package which implements openssl ffi bindings so that it is possible to use dtls in dart.

This PR is WIP to track the progress and signalize that i am working on it. Discussions on how to expose the option are welcome, too; currently it is only exposed via the `DefaultCoapConfig.coap` getter.

Implements #27

PR #69, opened 2022-03-28
refactor: add possibility to support different URI schemes

JKRhb:
Dealing with the addition of DTLS to the library, I noticed that the current structure of the library does not really allow for differentiating between URI schemes. This PR proposes a new structure that allows for adding additional network and channel types that can be chosen based on the URI scheme that is encountered.

This way, in contrast to what is currently proposed in #35, additional ways of transporting CoAP packets can be defined in their own classes and do not need to be integrated into the UDP channel/network classes. This also makes the library easier to extend with features like CoAP over TCP or CoAP over WebSockets.

Unfortunately, this PR requires #35 to be rebased/adjusted (sorry @Sorunome) but I hope it won't be too much work :/

### DIFF
--- a/lib/src/channel/coap_network_manager.dart
+++ b/lib/src/channel/coap_network_manager.dart
@@ -7,6 +7,18 @@
 
 part of coap;
 
+/// This [Exception] is thrown when an unsupported URI scheme is encountered.
+class UnsupportedProtocolException implements Exception {
+  /// The error message of this [Exception].
+  String get message => 'Unsupported URI scheme $uriScheme encountered.';
+
+  /// The unsupported Uri Scheme that was encountered.
+  final String uriScheme;
+
+  /// Constructor.
+  UnsupportedProtocolException(this.uriScheme);
+}
+
 /// A network management/caching class, allows already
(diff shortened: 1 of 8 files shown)


shamblett:
This looks OK to me, I take it it's now ready to merge, just checking.

JKRhb:
@shamblett  Thank you for your feedback :) Besides the merge conflict it would be ready, however, maybe the change could also be incorporated into @JosefWN's PR #76?

JosefWN:
I can incorporate it if you want and if @shamblett is ok with it, working on mutlicast and adding more examples/tests to find and work out any issues in the PR

shamblett:
No problem from me with this approach.

JosefWN:
Hope I didn't miss anything: https://github.com/shamblett/coap/pull/76/commits/4a44d63a7a76c0cac1d0a89cc60b1082d166307a :)

Also includes a fix for a formatting issue I had caused in the builder, which I discovered when I re-generated the CoapConfigs (which I did just in case).

JKRhb:
Awesome, thank you :) Then I'll close this PR :)

PR #71, opened 2022-03-29
Fix post example

JosefWN:
The post example was not using post. I change it to POST to `create-large`, used it for testing.

... also changed the put example from post to put.


JosefWN:
Pending https://github.com/shamblett/coap/pull/70, I have also removed the unnecessary `exit(0)` in the examples which were put in place to work around the bug that said PR fixes.

Issue #73, opened 2022-04-07
Package pedantic is discontinued


JosefWN:
See: https://pub.dev/packages/pedantic

Recommended alternative for Dart: https://pub.dev/packages/lints

shamblett:
Fixed on #74

PR #74, opened 2022-04-07
Increment message id, enable client polling

JosefWN:
Bugfixes:
1. Incrementing message id which the endpoint keeps track of, to prevent collisions preventing frequent polling

Minor fixes:
1. Default endpoint not re-created for every request, once an endpoint is set on a message it ensures the message id is set to the next available id
2. There was a `coapEndpoint` and an `endpoint` in `CoapClient`, switched to just keep a default `endpoint`
3. Removed superfluous `_method` in `CoapRequest` which was already held in its super class
4. No need to `resolveDestination`, if the `CoapRequest` has an endpoint, it has a destination
5. Fixed some minor linting issues

Suggestions:
1. Simplified token generation removing the need for `useRandomTokenStart` and the worry about wrapping around too fast. Assuming the random number generation is at least decently sound, even 100M requests would be less than 1% likely to collide on token, if my math is right.

Been running for 90 min now, about 4000 requests over 4 clients, 0 failures :)


JosefWN:
Ah, what the heck, fixed https://github.com/shamblett/coap/issues/73 in a separate commit while I was at it :P

JosefWN:
Ran a longer test yesterday, sent about 45 000 requests, not counting block requests, all of which were successful!

Only had this warning, which my latest commit should fix:
```
[+55325686 ms] flutter: 2022-04-07 22:14:58.976: WARNING: >> Truncating value {65536} to {16}-bit integer
[  +17 ms] flutter: 2022-04-07 22:14:59.025: WARNING: >> Matcher - Possible MID reuse before lifetime end: abe5fb343f3d3ccf expected MID
65536 but received 0
```

Issue #75, opened 2022-04-07
Library semantics


JosefWN:
A couple of things that puzzled me as I've dug into the code. I also want to stress that I understand that this is originally a port of another project, so I'm sure it carries some legacy from there, and I don't mean to nag or be a PITA. I've just created this issue to spark some discussion regarding possibilities revolving around the future direction of the repo.

1. The request _is_ the client (part of it, anyway), as opposed to being something which is _processed_ by the client.

This is not only confusing semantically but might also pose issues if this library was to support non-blocking sends in the future. Imagine:
```dart
  final request = CoapRequest.newGet()..addUriPath('large');
  client.request = request;
  client.get().then((response) => print(response.payloadString));

  final request2 = CoapRequest.newGet()..addUriPath('hello');
  client.request = request2;
  client.get().then((response) => print(response.payloadString));

  print(client.request); // Two ongoing requests, which one is `client.request`?
``` 

Makes more sense to:
```dart
  final request = CoapRequest.newGet()..addUriPath('large');
  client.send(request).then((response) => print(response.payloadString));

  final request2 = CoapRequest.newGet()..addUriPath('hello');
  client.send(request2).then((response) => print(response.payloadString));
```
I know you can do both (synchronously), but should you be able to? In a larger piece of code, even with synchronous sends, with `client.send(request)` it's clear what's going on without any added context, whereas with `client.get()` it's much less so.

**Proposal**: Drop `client.request` and related methods.

2. Same thing with the endpoint, I would have imagined a 1:1 mapping between the client and a socket/endpoint. If you want to send requests to other endpoints you create other clients.

Confusing example now:
```dart
  final conf = CoapConfig();
  final uri = Uri(scheme: 'coap', host: 'coap.me', port: conf.defaultPort);
  final client = CoapClient(uri, conf);

  final request = CoapRequest.newGet()
    ..addUriPath('large')
    ..endpoint = notCoapDotMe;
  var response = await client.send(request);
```

The client URI says [coap.me](https://coap.me) but the request is sent somewhere else.

**Proposal**: Make `request.endpoint` a private field, or drop it altogether if the refactoring is feasible. The request still has a `destination` via its inheritance from `CoapMessage`, and `CoapRequest send()` could perhaps be moved to the client. Clients typically send requests, requests don't send themselves.

3. The request is a request and a response combined:

```dart
  CoapResponse? _currentResponse;
  CoapResponse? get response => _currentResponse;
```

[...]

Issue #72, opened 2022-03-29
Multiple CoapConfig causes null pointer exception


JosefWN:
This limitation should perhaps at least be documented, better yet fixed for those who run multiple clients and want to use a base config as starting point.

Steps to reproduce:
```dart
FutureOr<void> main(List<String> args) async {
  // intended for client1
  final conf1 = CoapConfig();
  // intended for client2, default config but different port (example)
  final conf2 = CoapConfig()..defaultPort = 1234;

  final uri = Uri(scheme: 'coap', host: 'coap.me', port: conf1.defaultPort);
  final client1 = CoapClient(uri, conf1);

  final request = CoapRequest.newGet();
  request.addUriPath('large');
  client1.request = request;

  await client1.get();

  client1.close();
}
```

Output:
```
Unhandled exception:
Null check operator used on a null value
#0      CoapEndPoint._serializeRequest (package:coap/src/net/coap_endpoint.dart:230:28)
#1      CoapEndPoint.sendRequest (package:coap/src/net/coap_endpoint.dart:185:21)
#2      CoapStackBottomLayer.sendRequest (package:coap/src/stack/coap_layer_stack.dart:116:23)
#3      CoapNextLayer.sendRequest (package:coap/src/stack/coap_layer_stack.dart:20:10)
#4      CoapAbstractLayer.sendRequest (package:coap/src/stack/coap_abstract_layer.dart:18:15)
#5      CoapReliabilityLayer.sendRequest (package:coap/src/stack/coap_reliability_layer.dart:140:11)
#6      CoapNextLayer.sendRequest (package:coap/src/stack/coap_layer_stack.dart:20:10)
#7      CoapAbstractLayer.sendRequest (package:coap/src/stack/coap_abstract_layer.dart:18:15)
#8      CoapTokenLayer.sendRequest (package:coap/src/stack/coap_token_layer.dart:28:11)
#9      CoapNextLayer.sendRequest (package:coap/src/stack/coap_layer_stack.dart:20:10)
#10     CoapAbstractLayer.sendRequest (package:coap/src/stack/coap_abstract_layer.dart:18:15)
#11     CoapBlockwiseLayer.sendRequest (package:coap/src/stack/coap_blockwise_layer.dart:57:13)
#12     CoapNextLayer.sendRequest (package:coap/src/stack/coap_layer_stack.dart:20:10)
#13     CoapAbstractLayer.sendRequest (package:coap/src/stack/coap_abstract_layer.dart:18:15)
[...]

Issue #41, opened 2021-10-21
How is error handling supposed to work?


Sorunome:
I am having some trouble trying to grasp how error handling is supposed to work. If sending a message times out, `client.send()` returns and empty response, instead throwing a timeout exception.

When establishing a UDP connection fails, the error is cought (for logging), but never re-thrown. One could say that this is because your internet may be bad currently and it may be able to establish the connection later on, however it is never retried.

Maybe there is some fundamental concept of error in this package handling that i am missing?

shamblett:
AS the README says this package is a port of the C# .NET project [CoAP  .NET](https://github.com/smeshlink/CoAP.NET) so it pretty much follows what that did, if it threw an exception we will, if it logged we will etc. There are some differences caused by dart not supporting exactly what .NET does but in the main it should do the same.

If your asking is there an overall strategy of when we throw, log or retry then no.

JosefWN:
Yeah, +1 for consistency! I suppose it's a contentious topic, but in line with how standard libraries work, I typically only handle internal errors in my libraries (and perhaps debug logging at most). This leaves the user of the library to implement API-level error handling and logging at large, making it streamlined with the rest of their application.

shamblett:
Believe this has been addressed. Package re released at version 5.0.0.

PR #76, opened 2022-04-12
Proposal: rework top-level API

JosefWN:
This is an attempt to fix:
- https://github.com/shamblett/coap/issues/75
- https://github.com/shamblett/coap/issues/72
- https://github.com/shamblett/coap/issues/41

My work was loosely inspired by https://pub.dev/packages/http but with an adaptation to how CoAP clients commonly work in other languages (at least those I like 😅). See the examples to get a feeling for things. TLDR: this PR aims to make the library more native to Dart, to better leverage its strengths. Hopefully I'll manage to make a good case for it below...

Unsure:
- Removed the `validate` method on the client, I figured either the user builds his request (for example for etag validation, using `client.send`), or the user relies on the client for building the entire request (for example `client.get('hello')`). This felt like a bit of a hybrid.

Major stuff:
- There are request builders for the most common verbs (`get(...)`, `post(...)` etc.), a more manual `send(...)` for advanced requests and its counterpart `observe(...)` for streaming, which I suppose is not used often enough to warrant a request builder?
- Tightened up the API by exposing less internals and using the `meta` package with its `@protected` annotation for those I couldn't hide completely (will give static analysis warnings). It's now much harder for the user to "do the wrong things".
  - This also reduces the need for some of the error handling that was in place. In fact, we would have needed even more error handling with the old solution, there were plenty of ways to make the client misbehave/crash.
- Major removals:
  - Logging, which is not typically done in libraries (see Dart standard library for example). While the logging helps developers of this library debug, it complicates the code a fair bit, and is an obstacle for users of the library (in fact, all issues mentioned above are touching upon logging and error handling, directly or indirectly). If anything I would suggest adding more integration tests (examples) and working out as many bugs as possible that way. With a really stable API I don't think extensive logging is necessary. I could actually find the issues while working now, with a couple of temporary `print()` calls.
  - The CoAP.NET project hasn't seen active development since mid-2014 (8 years ago!), and keeping binaries in git is generally discouraged. It's also hard to use these tests on a non-Windows machine. In addition to coap.me, there is Californium, a project which seems to be alive and well, and whose plugtest-server you can run yourself cross-platform. If you miss any tests I propose adding more rather than bringing back the testserver directory, which also contained some duplicate code.
  - The flutter example project felt a bit awkward to keep in a Dart library, only more code to maintain. Flutter users will likely find using the library straight-forward anyhow (it was no problem for me at all). 
  - `CoapIChannel` and `CoapIDeliverer` and their implementations were seemingly not serving any purpose, even less so after I went from passing the response back to the request, to using the event bus to communicate. This in turn removes the need for the `CoapNetworkManager` since each client now only has one endpoint (less risk of API user mistakes), as well as the setting `poolUdpConnectionsByClient`. The pooling doesn't have any drawbacks as far as I can see, the more separation the better?
- Now supports async requests over a shared socket. This required separating different components to a greater degree, and relying less on global/static variables. Two major benefits:
1. The code doesn't crash on:
```dart
client.get('large').then((CoapResponse resp) => print(resp));
```
You can even run multiple observe and different requests concurrently (see `examples/get_observe_async.dart`).

2. Requests can be batched at far higher efficiency, example:
```bash
dart sync_vs_async.dart
Sending 10 async requests...
10 async requests took 49 ms
Sending 10 sync requests...
10 sync requests took 393 ms
```

Minor stuff:
- Fixes an issue with observe (no endpoint set, `_doPrepare` called one line too late)
- Fixes an issue where token is not set on the overarching `block1` requests (need it for matching in the client)
- Merged the `test/issues` folder with `examples`
- Removes `notify` and `error` on `observe`, they were unused arguments from what I could see
- Removes `raw_request_creation.dart` as that would no longer be supported (no need to worry about preparing the request manually)
- Many other small fixes...

Performance vs 3.5.0 (in parentheses), as baseline before all of my changes:
```
- 10 sync requests: 362 ms (410 ms)
- Ping: 93 ms (10033 ms) <- returning on reject now, not on timeout
- Blockwise (block2): 799 ms (846 ms) 
- Blockwise (block1): 248 ms (302 ms)
```

Nothing scientific, just a couple of quick checks to ensure my changes that went out in v3.6.0 and in this PR are not making things slower.


JosefWN:
Please don't merge yet, but feel free to come with input! I'm thinking of moving `send` and `waitForResponse` into the client in a separate commit if it pans out well, which can also be reverted if you guys don't like it, will do it tonight or tomorrow.

JosefWN:
Ok! Ready for review @shamblett, @JKRhb! Linting, formatting & tests are passing, examples are working :)

Not sure how to go about reviewing what turned out to be a gargantuan PR, sorry for that. Perhaps the best is to just start with viewing and running the examples to get a feel for the API. Everything is tested except for multicast, didn't have a good way of testing it.

JKRhb:
Thank you for your hard work, @JosefWN!

> Not sure how to go about reviewing what turned out to be a gargantuan PR, sorry for that.

Maybe aspects like the test server and flutter example removal could be spun off into its own PR? That would reduce the size of the PR a bit

examples/multi_client.dart
```diff
@@ -0,0 +1,41 @@
...
+      port: conf2.defaultPort);
+  final client2 = CoapClient(uri2, conf2);
+
+  try {
+    print('Sending get /hello to ${uri1.host}');
+    var response = await client1.get('hello');
+    print('/hello response: ${response.payloadString}');
+
+    print('Sending get /test to ${uri2.host}');
+    response = await client1.get('test');
```
JKRhb, line 32: I noticed one typo here:

```suggestion
    response = await client2.get('test');
```

lib/src/network/coap_network_udp.dart
```diff
@@ -0,0 +1,95 @@
...
+                  CoapInternetAddress(d.address.type, d.address);
+              _eventBus.fire(CoapDataReceivedEvent(buff, coapAddress));
+            }
+          }
+          break;
+        case RawSocketEvent.closed:
+        case RawSocketEvent.readClosed:
+          close();
+          break;
+        case RawSocketEvent.write:
```
JKRhb, line 74: I noticed the linter is complaining a bit here:

```suggestion
        default:
          break;
```

lib/src/coap_request.dart
```diff
@@ -18,28 +18,28 @@ class CoapRequest extends CoapMessage {
...
+  CoapRequest.withType(int code) : this.isConfirmable(code, confirmable: true);
 
   /// Initializes a request message.
   /// True if the request is Confirmable
-  CoapRequest.isConfirmable(int? code, {required bool confirmable})
-      : super.withCode(
-            confirmable ? CoapMessageType.con : CoapMessageType.non, code);
+  CoapRequest.isConfirmable(int code, {required bool confirmable})
+      : super.withCode(code,
+            type: confirmable ? CoapMessageType.con : CoapMessageType.non);
```
JKRhb, line 27:
```dart
class CoapRequest extends CoapMessage {
  /// Initializes a request message.
  /// Defaults to confirmable
  CoapRequest(int code, {confirmable = true})
      : super(
            code: code,
            type: confirmable ? CoapMessageType.con : CoapMessageType.non);

```

lib/src/coap_message.dart
```diff
@@ -18,10 +18,10 @@ class CoapMessage {
   CoapMessage();
 
   /// Instantiates a message with the given type.
-  CoapMessage.withType(this.type);
+  CoapMessage.withType(this.type, {this.code = CoapCode.notSet});
 
   /// Instantiates a message with the given type and code.
-  CoapMessage.withCode(this.type, this.code);
+  CoapMessage.withCode(this.code, {this.type = CoapMessageType.unknown});
```
JKRhb, line 24:
```dart
class CoapMessage {
  /// Constructor
  CoapMessage(
      {this.type = CoapMessageType.unknown, this.code = CoapCode.notSet});
```

JKRhb (commented):
I think both the `CoapMessage` and the `CoapRequest` constructors could be simplified a bit. (This requires some more changes at other places in the code base.)

examples/multi_client.dart
JosefWN: Ah, whoops, fixed: https://github.com/shamblett/coap/pull/76/commits/c62836d6601d0f0800884d4125628e452cb5502e

lib/src/network/coap_network_udp.dart
JosefWN: Fixed: https://github.com/shamblett/coap/pull/76/commits/c62836d6601d0f0800884d4125628e452cb5502e

lib/src/coap_request.dart
JosefWN: Much cleaner! Changed: https://github.com/shamblett/coap/pull/76/commits/c62836d6601d0f0800884d4125628e452cb5502e

lib/src/coap_message.dart
JosefWN: Agreed: https://github.com/shamblett/coap/pull/76/commits/c62836d6601d0f0800884d4125628e452cb5502e

JosefWN:
> Maybe aspects like the test server and flutter example removal could be spun off into its own PR? That would reduce the size of the PR a bit

Yeah, the reason I included them in the same PR was because I didn't want to invest time in cleaning them up and fixing potential changes causing them to break on master. Going out now but I can re-add them tomorrow and then make a new PR if/once this one gets merged.

JosefWN:
@JKRhb done, also reduced duplicate code in the client a bit :)

lib/src/net/coap_exchange.dart
```diff
@@ -37,7 +37,7 @@ class CoapExchange {
   /// The current request
   CoapRequest? currentRequest;
 
-  /// The response
+  // The response
```
JKRhb, line 40: Hmm, is it intentional that this is not a doc comment anymore?

lib/src/net/coap_iendpoint.dart
```diff
@@ -15,23 +15,20 @@ abstract class CoapIEndPoint {
...
-  /// Gets or sets the message deliverer.
-  CoapIMessageDeliverer? deliverer;
-
   /// Gets the outbox.
   CoapIOutbox get outbox;
 
   /// Starts this endpoint and all its components.
   Future<void> start();
 
+  // The namespace which the endpoint belongs to
```
JKRhb, line 29:
```suggestion
  /// The namespace which the endpoint belongs to
```

lib/src/observe/coap_observe_relation.dart
```diff
@@ -14,15 +14,16 @@ class CoapObserveRelation {
   /// The observing endpoint
```
JKRhb, line 14: I think this file could probably be adjusted a bit:

```dart
/*
 * Package : Coap
 * Author : S. Hamblett <steve.hamblett@linux.com>
 * Date   : 31/10/2017
 * Copyright :  S.Hamblett
 */

part of coap;

/// Represents a relation between a client endpoint and a resource on this
/// server.
class CoapObserveRelation {
  /// Constructs a new observe relation.
  ///
  /// Takes the observing [endpoint], the observed [resource], and the
  /// [exchange] that tries to establish the observe relation as arguments.
  CoapObserveRelation(this.cfg, CoapObservingEndpoint endpoint,
      CoapIResource resource, CoapExchange exchange)
      : _endpoint = endpoint,
        _resource = resource,
        _exchange = exchange,
        established = true;

  DefaultCoapConfig cfg;

  final CoapObservingEndpoint _endpoint;

  /// Source endpoint of the observing endpoint
  InternetAddress? get source => _endpoint.endpoint;
  final CoapIResource _resource;

  /// The resource
  CoapIResource? get resource => _resource;
  final CoapExchange _exchange;

  /// The exchange
  CoapExchange get exchange => _exchange;

  /// Current control notification
  CoapResponse? currentControlNotification;

  /// Next control notification
  CoapResponse? nextControlNotification;

  /// Key
  String get key => '$source#${_exchange.request?.tokenString}';

  /// A value indicating if this relation has been established
  bool established;
  DateTime _interestCheckTime = DateTime.now();
  int _interestCheckCounter = 1;

  /// The notifications that have been sent, so they can be
  /// removed from the Matcher.
  final Queue<CoapResponse?> _notifications = Queue<CoapResponse?>();

  /// Cancel this observe relation.
  void cancel() {
    // Stop ongoing retransmissions
    if (_exchange.response != null) {
      _exchange.response!.isCancelled = true;
    }
    established = false;
    _resource.removeObserveRelation(this);
    _endpoint.removeObserveRelation(this);
    _exchange.complete = true;
  }

  /// Cancel all observer relations that this server has
  /// established with this's realtion's endpoint.
  void cancelAll() {
    _endpoint.cancelAll();
  }

  /// Notifies the observing endpoint that the resource has been changed.
  void notifyObservers() {
    // Makes the resource process the same request again
    _resource.handleRequest(_exchange);
  }

  /// Check
  bool check() {
    var check = false;
    final now = DateTime.now();
    check = check ||
        _interestCheckTime
            .add(Duration(milliseconds: cfg.notificationCheckIntervalTime))
            .isBefore(now);
    check = check ||
        (++_interestCheckCounter >= cfg.notificationCheckIntervalCount);
    if (check) {
      _interestCheckTime = now;
      _interestCheckCounter = 0;
    }
    return check;
  }

  /// Add a notification
  void addNotification(CoapResponse notification) {
    _notifications.add(notification);
  }

  /// Clear notifications
  Iterable<CoapResponse?> clearNotifications() {
    Iterable<CoapResponse?> list;
    list = _notifications.toList();
    _notifications.clear();
    return list;
  }
}
```

lib/src/observe/coap_observe_relation.dart
```diff
@@ -14,15 +14,16 @@ class CoapObserveRelation {
...
+  CoapObserveRelation(this.cfg, CoapObservingEndpoint endpoint,
+      CoapIResource resource, CoapExchange exchange) {
     _endpoint = endpoint;
     _resource = resource;
     _exchange = exchange;
     _key = '$source#${exchange.request!.tokenString}';
   }
 
-  final CoapILogger? _log = CoapLogManager().logger;
+  DefaultCoapConfig cfg;
```
JKRhb, line 25: `cfg` could be renamed to `config`

lib/src/stack/coap_blockwise_layer.dart
```diff
@@ -286,45 +254,37 @@ class CoapBlockwiseLayer extends CoapAbstractLayer {
...
     }
 
     final block2 = response.block2;
     if (block2 != null) {
-      _log!.info('Blockwise block2 - response acknowledges block: $block2');
       var status = _findResponseBlockStatus(exchange, response);
-      _log!.info('Blockwise exchange block2 status is - $status');
       final blockStatus = CoapBlockOption(optionTypeBlock2);
-      blockStatus.rawValue = status.currentNUM;
+      blockStatus.rawValue = status.currentNUM!;
```
JKRhb, line 264: I think the type of `currentNUM` could be changed from `int?` to `int`.

lib/src/stack/coap_chain.dart
```diff
@@ -265,17 +265,17 @@ class CoapChain<TChain, TFilter, TNextFilter>
       }
       e = e.nextEntry as CoapEntry<TFilter?, TNextFilter>?;
     }
-    return e as CoapIEntry<TFilter, TNextFilter>;
+    return e! as CoapIEntry<TFilter, TNextFilter>;
```
JKRhb, line 268: I think the `!` could be removed here

```suggestion
    return e as CoapIEntry<TFilter, TNextFilter>;
```

lib/src/stack/coap_chain.dart
```diff
@@ -265,17 +265,17 @@ class CoapChain<TChain, TFilter, TNextFilter>
...
   @override
   CoapIEntry<TFilter, TNextFilter> getEntryByType(Object filterType) {
-    var e = _head!.nextEntry as CoapEntry<TFilter, TNextFilter>;
+    var e = _head!.nextEntry! as CoapEntry<TFilter, TNextFilter>;
     while (e != _tail) {
       if (filterType is TFilter) {
         return e;
       }
-      e = e.nextEntry as CoapEntry<TFilter, TNextFilter>;
+      e = e.nextEntry! as CoapEntry<TFilter, TNextFilter>;
```
JKRhb, line 278: Same here I think.

lib/src/stack/coap_chain.dart
```diff
@@ -421,7 +421,7 @@ class CoapChain<TChain, TFilter, TNextFilter>
   }
 
   CoapEntry<dynamic, dynamic> _checkOldName(String baseName) =>
-      _name2entry[baseName] as CoapEntry<dynamic, dynamic>;
+      _name2entry[baseName]! as CoapEntry<dynamic, dynamic>;
```
JKRhb, line 424: I think this null check operator is also not needed.

JKRhb (commented):
I think the PR looks pretty good, see a last couple of comments/suggestions below :)

lib/src/stack/coap_chain.dart
JosefWN: No: https://dart-lang.github.io/linter/lints/cast_nullable_to_non_nullable.html

JKRhb: Oh, I see, thanks for the clarification!

lib/src/stack/coap_chain.dart
JosefWN: I think in general it's better to be explicit in the code. In this case the cost is also only one character to clarify that the variable can be null, but that we are not expecting it to be?

lib/src/coap_client.dart
```diff
@@ -35,338 +55,353 @@ class CoapClient {
...
 
-  /// Let the client use early negotiation for the blocksize
-  /// (16, 32, 64, 128, 256, 512, or 1024). Other values will
-  /// be matched to the closest logarithm.
-  CoapClient useEarlyNegotiation(int size) {
-    _blockwise = size;
-    return this;
-  }
+  /// The default request timeout
+  late Duration timeout; // ms
```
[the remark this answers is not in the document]
JosefWN: Ah, we should remove the milliseconds thing altogether in the comments there, since now you can just set `client.timeout = Duration(seconds: 2)` for example, which makes it explicit to the library user anyway. The unit is only important for the `CoapConfig` now.

lib/builder.dart
```diff
@@ -54,34 +54,30 @@ import 'package:coap/coap.dart';
...
-  @override
-  CoapISpec? spec;
 ${_generateDataScript(data)}}
 """;
 
 String _generateDataScript(YamlMap data) {
   final buff = StringBuffer();
-  data.forEach((k, v) {
-    buff.writeln('');
+  for (var k in data.keys) {
```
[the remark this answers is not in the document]
JosefWN: Yeah, in general I was planning on making a pass over the code checking for that as well. There are numerous places in the code where we could be using final.

We could also consider adding some basic linter rules, for example:
```yaml
linter:
  rules:
    - prefer_final_fields
    - prefer_final_in_for_each
    - prefer_final_locals
    - prefer_final_parameters
    - prefer_const_constructors
    - prefer_const_constructors_in_immutables
    - prefer_const_declarations
    - prefer_const_literals_to_create_immutables
```

The base linter config is pretty lenient.

lib/src/stack/coap_chain.dart
JKRhb: Yeah, also considering your last comment, it is certainly better to leave it as for now :)

lib/src/net/coap_exchange.dart
JosefWN: Not intentional, fixed: https://github.com/shamblett/coap/pull/76/commits/bf2e194af342174accb5b8c413397e7d3f129cdf

lib/src/net/coap_iendpoint.dart
JosefWN: Fixed: https://github.com/shamblett/coap/pull/76/commits/bf2e194af342174accb5b8c413397e7d3f129cdf

lib/src/observe/coap_observe_relation.dart
JosefWN: Fixed: https://github.com/shamblett/coap/pull/76/commits/ad10f99673d9303bd2709d6c05a78c065909d09b

lib/src/stack/coap_chain.dart
JosefWN: Ok! Marking this as resolved now then :)

lib/src/coap_client.dart
```diff
@@ -35,338 +55,353 @@ class CoapClient {
...
+    CoapMulticastResponseHandler? onMulticastResponse,
+  }) {
+    final request = CoapRequest.newGet();
+    _build(request, path, accept, type, options, block2Size, maxRetransmit);
+    return send(request,
+        timeout: timeout, onMulticastResponse: onMulticastResponse);
+  }
+
+  /// Sends a POST request.
+  /// block2Size: early negotiation blocksize (16, 32, 64, 128, 256, 512, or 1024)
```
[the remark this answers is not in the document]
JosefWN: Yeah, I've seen enum used in other libs for SZX, just like it is specified in the RFC. Makes it more transparent, now the payload size will be rounded and unless the user has paid attention to the doc comment they might be a bit surprised. I can change to an enum and have a look at a block1 setting as well.

lib/src/coap_client.dart
```diff
@@ -35,338 +55,353 @@ class CoapClient {
...
-  /// The current request being processed by the client.
-  CoapRequest? request;
-
-  /// Timeout
-  int timeout = 32767; //ms
+  CoapClient(this.uri, this._config,
+      {this.addressType = InternetAddressType.IPv4, Duration? timeout}) {
+    this.timeout = timeout ?? Duration(milliseconds: _config.defaultTimeout);
+    _namespace = hashCode.toString();
+  }
```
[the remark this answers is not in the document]
JosefWN: Yes, I don't know how we should treat the config. I figured it could make more sense to move our non-RFC timeout default (which we had apparently set to 32767 ms) there as well, so we have a single config. Then again, we already have an implicit total timeout specified in the config, so should we define yet another timeout for the client like we do now? Regardless, the config will likely span multiple RFC's (7252 for UDP, 8232 for TCP, 7641 for observe, 7959 for blockwise), so it will probably not map 1:1 to RFC 7252.

lib/src/coap_client.dart
```diff
@@ -35,338 +55,353 @@ class CoapClient {
...
+  late Duration timeout; // ms
 
-  /// Let the client use late negotiation for the block size (default).
-  CoapClient useLateNegotiation() {
-    _blockwise = 0;
-    return this;
-  }
+  final DefaultCoapConfig _config;
+  CoapIEndPoint? _endpoint;
+  late final String _namespace;
```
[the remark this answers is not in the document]
JosefWN: Yeah, the reason I did it this way was to avoid repeated type conversions, but I can change, I'm not strongly in favor of either approach and it's a tiny cost.

lib/src/coap_client.dart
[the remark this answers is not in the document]
JosefWN: Fixed: https://github.com/shamblett/coap/pull/76/commits/23f12832be6d414080a70c4193c1a4afe34c44e9

lib/src/coap_client.dart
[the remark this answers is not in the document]
JosefWN: Fixed: https://github.com/shamblett/coap/pull/76/commits/23f12832be6d414080a70c4193c1a4afe34c44e9

lib/src/coap_client.dart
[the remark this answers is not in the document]
JosefWN: Fixed the constructor: https://github.com/shamblett/coap/pull/76/commits/23f12832be6d414080a70c4193c1a4afe34c44e9

If you want I can remove `defaultTimeout` from the config and add a separate issue to discuss whether we should specify the timeout twice?

lib/src/observe/coap_observe_relation.dart
JosefWN: Fixed: https://github.com/shamblett/coap/pull/76/commits/806d2aabc363905fe1dde8efcbf881d42f3118a3

lib/src/coap_client.dart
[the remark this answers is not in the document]
JKRhb: > Fixed the constructor: [23f1283](https://github.com/shamblett/coap/commit/23f12832be6d414080a70c4193c1a4afe34c44e9)
> 
> If you want I can remove `defaultTimeout` from the config and add a separate issue to discuss whether we should specify the timeout twice?

That sounds like a good idea :)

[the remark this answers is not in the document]
JosefWN: Fixed: https://github.com/shamblett/coap/pull/76/commits/175151b7968921eae2d53806b700e69a345a19d5

lib/src/stack/coap_blockwise_layer.dart
JosefWN: Fixed: https://github.com/shamblett/coap/pull/76/commits/e3d2bff4bf1ebcf93e8a832e13d29aa9c4a2f3f7

lib/src/coap_client.dart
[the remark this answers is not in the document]
JosefWN: Also I think all block1 requests use early negotiation always (since the first block will contain data anyway), but I could be wrong: https://datatracker.ietf.org/doc/html/rfc7959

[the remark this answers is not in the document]
JKRhb: > I suggest holding off until we have enhanced enums in Dart 2.17, created issue: https://github.com/shamblett/coap/issues/80

That sounds reasonable :)

> Also I think all block1 requests use early negotiation always (since the first block will contain data anyway), but I could be wrong: https://datatracker.ietf.org/doc/html/rfc7959

Hmm, I think by indicating the block1 value you could set the client's preferred/initial block size (c. f. figure 9). Having another look into the source code, though, I am not sure if "manually" adding a block1 option/setting the size would actually be processed at the moment. Maybe this is another potential issue that can be addressed in a follow-up PR :)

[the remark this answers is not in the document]
JosefWN: Yeah, but what makes the negotiation different is that for every block in a block1 request we set:
```dart
    block
        .addOption(CoapBlockOption.fromParts(optionTypeBlock1, num, szx, m: m));
```

This includes the first request/block. There will still be a negotiation, where the block1 size in subsequent messages is adjusted to the smallest of the client's preferred block1 size and the server's maximum block1 size.

In a block2 get for example, the get request itself is not blockwise, so we can negotiate early by including the block2 options as an instruction for the blockwise response from the server.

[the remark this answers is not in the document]
JosefWN: But in general I'm wondering if we should include this option to alter block size for specific requests at all, I just kept it because it was exposed to the user in the first place.

The block size is also set depending on connectivity (likelihood of IP packet fragmentation and such), so it would make sense to use the same settings for the whole client since it now maps 1:1 with a specific server, and this is specified in the config already. If you want to push larger payloads to the client you wait longer, you typically don't increase the block size.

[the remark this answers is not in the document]
JosefWN: Example: https://github.com/shamblett/coap/pull/76/commits/8dc6fb302e08398a7324e25ab3966f75806e0d66

So there is no `defaultBlockSize` but a `preferredBlockSize` as per the RFC. This is what the client as a whole prefers. Then you can choose whether or not to instruct the server to use the client's preference in its block2 response.

The whole negotiation is "piggybacked" with the requests/responses, so for block1 you can't negotiate early unless you want an extra roundtrip (very expensive), because the first block1 request already contains a payload.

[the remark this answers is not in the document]
JKRhb: Oh, that is a very elegant solution :) Very cool, thank you! :)

JKRhb (approved):
LGTM :) @shamblett Can you also have a look on this PR? :)

lib/src/coap_client.dart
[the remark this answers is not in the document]
JosefWN: Thank you, too, for focusing on the uglier parts of my refactor to make it better :)

lib/src/net/coap_exchange.dart
```diff
@@ -219,7 +227,7 @@ class CoapKeyToken {
 class CoapKeyUri {
   /// Construction
   CoapKeyUri(this._uri, this._endpoint) {
-    _hash = _uri.hashCode * 31 + (_endpoint == null ? 0 : _endpoint.hashCode);
+    _hash = _uri.hashCode * 31 + (_endpoint?.hashCode ?? 0);
```
JKRhb, line 230: I guess here we could use something like this?

```suggestion
    _hash = Object.hash(_uri, _endpoint);
```

lib/src/coap_option.dart
```diff
@@ -179,8 +179,7 @@ class CoapOption {
   @override
   int get hashCode {
     const prime = 31;
-    var result = 1;
-    return result = prime * result + _type;
+    return prime + _type;
   }
```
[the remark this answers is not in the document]
JosefWN: Ah, nice catch. In addition, if all fields are included in `Object.hash()`, I don't think we need to override `hashCode` or `operator ==`?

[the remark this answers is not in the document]
JosefWN: Okay, I got found a couple more places to clean up while I was at it: https://github.com/shamblett/coap/pull/76/commits/4cf829c056e6a63e408a3e93d991d548a413952f

With client mapping 1:1 to a socket we don't need `CoapKeyUri`, we can match blockwise requests on token. Having a `List<Map<int?, CoapExchange>>` for matching on id and a `Map<String, CoapExchange>` for matching on token works just fine. There is no need for the other two wrapping classes.

Re-ran all examples and everything works.

shamblett:
I've merged this to its own branch for now, issue75, while I check it out.

One question about logging, the original logging was really supplied so users could just turn on logging and supply me with a log of what happened when they raised an issue, without this we are pretty blind. The old logging although as you say problematic in itself did allow post mortem debugging for issue investigation. Are we now saying we are just replacing the old logging with print statements? This is OK but we mustn't lose detail here, the examples now provide minimal logging(OK these of course work so that maybe why). 

Will we still be able to see message detail, i.e token's/sequence numbers/flags/block counts etc. that we used to do?
These may be invaluable,especially in cases where the CoAP server the user is using doesn't work properly(we have had this in the past). We may not be able to reproduce the fault, nor give enough detail to the user as to what the problem is.

Sorry if I've missed something here.

JosefWN:
Hi @shamblett, and sorry for the large PR!

Personally I'm a big fan of the event bus since it helps us not only decouple the components within the library, but also gives us the opportunity to introspect the request/response flow. In this PR I always fire events (even if there is a hook) for this purpose. I'm also always including data with all events, so we can differentiate between multiple concurrent requests.

To make debug logging easier, we could expose the bus on the client as well, and add `toString` methods on the events. With that in place, I think what you are asking for could be achieved with this one-liner:
```dart
client.events.on().listen(print);
```

Example: https://github.com/JosefWN/coap/commit/087aafb43827b91d675001f7b15fd5d10b50dbc1

If you try running the `log_events.dart` example I included in the commit, you will see all blockwise activity.

This also makes it possible for library users to listen for and act on any events they choose, without imposing a specific logger or event handler on them (of course, we would still handle the events internally as required, but we wouldn't programmatically obscure anything from the library user). If we are missing important events for debugging we can just add more of them. What do you think?

Another thing we could do when users encounter issues is to ask them to provide us with a minimal example reproducing that issue. Shifting the burden of proof to the person with the problem will reduce the debugging load on us even further, as we could just run their problematic example code, or a very similar piece of code, against another server.

shamblett:
This sounds good to me, the existing logging did need to go, I'll have a look at log_events.dart but I'm happy with this approach.

PR #83, opened 2022-04-29
Add logging example

JosefWN:
Adds the new logging adjustments and example.


shamblett:
Yep I like this, so much so I may use this technique in some of my other packages. 

I'll carry on looking at issue 75 now thanks.

PR #86, opened 2022-05-02
Add DTLS support based on `dtls` and `dart_tinydtls`

### LINKED ISSUE #27: DTLS support for coap libary
May it is possible to build up the dtls support fot the coap-client

JKRhb:
This PR adds experimental DTLS support to the project based on a wrapper around [Eclipse tinydtls](https://github.com/eclipse/tinydtls/tree/develop), supporting Pre-Shared Keys as well as Raw Public Keys. The config is adjusted in a way that allows adding other DTLS backends like OpenSSL (see #35) in the future.

I also included an example that illustrates the usage of CoAPS with a Pre-Shared Key with the Californium test server. The example can be run on both Linux and Windows.

Edit: Now resolves #35 (as I included the addition of OpenSSL support by @Sorunome in the meantime), closes #27.


JKRhb:
Converted the PR back to a draft as I noticed a couple of breaking changes in `dart_tinydtls` that should be integrated first.

lib/src/network/coap_inetwork.dart
```diff
@@ -50,14 +50,40 @@ abstract class CoapINetwork {
...
+            namespace: namespace,
+            pskCredentials: pskCredentials,
+            ecdsaKeys: ecdsaKeys);
       default:
         throw UnsupportedProtocolException(uri.scheme);
     }
   }
+
+  /// Determines which [DtlsBackEnd] to use for secure communication.
+  static CoapINetwork _determineDtlsBackend(
```
JosefWN, line 72: Should we support multiple backends or should we pick the one we think is best and stick with that? Would make the code a bit simpler, or we are worried about platform support? It's a shame there are only bindings in Dart for DTLS from what I could see, nothing native :(

lib/src/coap_client.dart
```diff
@@ -88,6 +88,9 @@ class CoapClient {
   CoapIEndPoint? _endpoint;
   final _lock = sync.Lock();
 
+  /// Raw Public Keys for CoAPS with tinyDtls.
+  EcdsaKeys? ecdsaKeys;
```
JosefWN, line 92: Probably good if `EcdsaKeys` and `PskCredentials` are implemented as our own types, such as `CoapEcdsaKeys` and `CoapPskCredentials` if it's feasible? That way we can switch the DTLS backend without affecting the user-facing API in the future, if at some point there is a native DTLS library (or SDK support).

lib/src/network/coap_inetwork.dart
JKRhb: Tinydtls does not support certificates, while OpenSSL does not support Raw Public Keys, yet :/ The OpenSSL binding by @Sorunome et al. is currently also missing support for Pre-Shared Keys (I tried to [integrate the feature](https://gitlab.com/famedly/company/frontend/libraries/dtls/-/merge_requests/2) in the upstream, but the project seems dormant right now). So both have features the other one is missing, which is why I thought about having multiple backends in the first place.

What I had in my mind was having both the tinydtls backend and the OpenSSL one for a certain period of time in parallel in order to cover all CoAPS features and then gradually replace them once native DTLS support becomes available. I actually also started experimenting a bit with integrating DTLS into the Dart SDK itself (which would most definitely be the cleanest solution), but figured that it might take some more time to get it into the upstream. The integration of tinydtls would therefore be a bit of a temporary workaround that hopefully can be replaced in the not too distant future.

lib/src/coap_client.dart
JKRhb: Sounds good, I will update the PR accordingly :)

JosefWN:
Would be nice if we included instructions for how to download or build the binaries, instead of including them. Keeping binaries in git is generally considered a bad practice unless absolutely necessary. Binary diffs can be large, so not only do they grow the repo on first commit, but even more so over time.

Binaries also makes the releases significantly larger. A packaged release is about 600 KB compressed now (v3.6.0), if we include binaries for all platforms and architectures supported by Dart the DTLS examples will be much larger than the entire repo . Even the binaries included in this PR (only Windows and Linux x86?) weigh in at almost the size of the rest of the repo and all dependencies combined, which is quite a lot considering everyone might not be using DTLS, or the platforms currently supported by the examples (personally I work mostly with Mac/x86+ARM and iOS/ARM).

lib/src/coap_client.dart
JKRhb: I reimplemented `EcdsaKeys` and `PskCredentials` (as well as a new `PskCredentialsCallback`) now within this library. The contents of those classes are then mapped to the tinydtls specific definitions. (I changed the API in the tinydtls wrapper to allow for handling PSK identity hints, which hasn't been possible before.)

JKRhb:
Regarding the binaries: I updated the README now a bit, indicating that you can download them from GitHub. This is still far from optimal, but probably better than before.

lib/src/network/coap_inetwork.dart
JKRhb: I now removed the backend logic and let the library use tinydtls if a PSK or ECDSA keys are provided. This way, users don't need to bother with the internals and the API can probably be kept stable, even if the backend is swapped out or another one for certificates is added.

JKRhb: Maybe the method's name could now be changed to `_determineDtlsNetwork`?

JKRhb: I renamed the method now to `_determineDtlsNetwork´ :)

JKRhb:
I now also integrated #35 by @Sorunome in this PR :)

lib/builder.dart
```diff
@@ -54,30 +54,39 @@ import 'package:coap/coap.dart';
...
+      buff.writeln('');
+    }
     buff.writeln('  @override');
-    if (data[k] is String) {
-      if (data[k] == 'true' || data[k] == 'false') {
-        buff.writeln('  bool get $k => ${data[k]};');
-        continue;
+    firstLine = false;
+    if (v is String) {
+      if ('true' == v || 'false' == v) {
```
JosefWN, line 75: Tiny nitpick: I would reverse these statements, so they read `v == 'true' || v == 'false'`, putting the variable is on the left-hand side and the value it's compared to on the right-hand side.

It reads a bit easier and is in line with the rest of the code (`v == null` as opposed to `null == v`, `v is bool` as opposed to `bool is v` etc.).

lib/src/coap_client.dart
```diff
@@ -364,7 +379,9 @@ class CoapClient {
         final socket = CoapINetwork.fromUri(uri,
             address: destination,
             config: _config,
-            namespace: _eventBus.namespace);
+            namespace: _eventBus.namespace,
+            ecdsaKeys: ecdsaKeys,
+            pskCredentialsCallback: request.pskCredentialsCallback);
```
JosefWN, line 384: Since the credentials callback can be set on every request, and the endpoint is re-used (and potentially used concurrently by multiple requests), perhaps we need to think of another way of passing the `pskCredentialsCallback`.

Do we expect it to change from one request to another or could it be set at another level, for example as an optional parameter passed to the client itself?

lib/src/coap_config.dart
```diff
@@ -83,4 +83,24 @@ abstract class DefaultCoapConfig {
   int get markAndSweepInterval => 10 * 1000; // ms
 
   int get channelReceivePacketSize => 2048;
+
+  /// Whether tinydtls bindings via the [tinydtls] package should be used for
+  /// CoAPS.
+  bool get dtlsUseTinyDtls => false;
```
JosefWN, line 89: Could we perhaps handle this similar to how version is handled? For example:
```dart
String get dtlsBackend => 'TinyDtls';
// or
String get dtlsBackend => 'OpenSSL';
```

If I got it right, the user should select one DTLS backend. What should the user expect if `dtlsUseTinyDtls` and `dtlsUseOpenSSL` are both set to `true`? It would be a stupid config (unless I misunderstood), but if the config itself prevents that ambiguity it could make things a little clearer to the end user.

JKRhb: Oh, I actually intended both to be usable at the same time as they have different feature sets. Once the OpenSSL binding catches up, tinydtls could then be phased out gradually.

lib/src/coap_client.dart
JKRhb: Setting it at the client level would certainly be an option, especially since connections are reused between requests.

lib/src/coap_config.dart
JosefWN: But doesn't the user ultimately have to choose one of the two available backends to use with one specific client? Just took a quick glance so I could be wrong, I was just judging by `_determineDtlsNetwork`, which seems to return only a `CoapNetworkTinyDtls` if both TinyDtls and OpenSSL are used at the same time?

I understand that you might want to have one client using TinyDtls and another client using OpenSSL, in which case their client configurations would be different, but using two backends simultaneously to send the same requests is confusing me a bit.

JKRhb: Hmm, yeah, the logic is a bit complicated and not well documented, yet. What I intended is the following:

- Assume the user has both tinydtls and OpenSSL enabled. If the user does not provide PSK or ECDSA credentials, then OpenSSL (i.e., certificate mode) is used. If at least one of the two kinds of credentials is present, then tinydtls is used and OpenSSL will be ignored.
- In the case that only tinydtls is loaded, and no credentials are provided, an exception is thrown.
- If only OpenSSL is active, then the client will always use certificates.
- Finally, if no backend is present, another kind of exception is thrown.

So, yes, the user does in fact choose a backend, but only implicitly by either providing credentials or not (after the initial review I thought this might be better to have a more stable API). However, if this should be made more explicit, then I would reintroduce an enum (or a string?) to have users make a conscious decision in this regard.

lib/builder.dart
JKRhb: Made some adjustments that should make the code a bit more readable :)

lib/src/coap_config.dart
JosefWN: Ah, I understand. I was thinking the developer would consciously decide which library to bundle based on the library's capabilities, since there is some effort involved in setting up and maintaining the bundling. I think OpenSSL also has to be bundled for some platforms (iOS/Android), common targets for Flutter which also happen to be on metered connections where a smaller app size is an advantage. Not sure about Windows, but I think OpenSSL is not shipped with the OS there either?

Might be better to be explicit in the config (with an enum or such), since at the end of the day, you will typically only be using one or the other? Then the user will also know which backend is used, and will not feel inclined to needlessly bundle both.

But I'm not of a strong opinion, this is likely only necessary during a transitional period, right? Can't imagine Dart going on forever without decent DTLS support :P

JKRhb: > I think OpenSSL also has to be bundled for some platforms (iOS/Android), common targets for Flutter which also happen to be on metered connections where a smaller app size is an advantage. Not sure about Windows, but I think OpenSSL is not shipped with the OS there either?

Oh, yeah, that is a good point! And yes, OpenSSL has to be explicitly installed on Windows, if I am not mistaken.

> Might be better to be explicit in the config (with an enum or such), since at the end of the day, you will typically only be using one or the other? Then the user will also know which backend is used, and will not feel inclined to needlessly bundle both.

Sounds good, I updated the PR accordingly :)

> But I'm not of a strong opinion, this is likely only necessary during a transitional period, right? Can't imagine Dart going on forever without decent DTLS support :P

I really hope so. It is pretty interesting that DTLS seems to be such a challenge for many ecosystems (in Node.js, adding DTLS support to Node core has been discussed for years with no significant progress). I think adding it to the Dart SDK should be doable, though :) Especially with the work already done with the two backends present in this PR.

lib/src/coap_client.dart
JKRhb: Moved the callback to the client in the meantime :)

lib/builder.dart
```diff
@@ -54,30 +54,58 @@ import 'package:coap/coap.dart';
...
+
+  data.forEach((key, value) {
+    if (value is YamlMap) {
+      buff.write(_generateDataScript(value, key, firstLine: false));
+      return;
+    }
+    final variableName =
+        prefix.isEmpty ? key : prefix + key[0].toUpperCase() + key.substring(1);
+
+    if (variableName == 'dtlsBackend') {
```
JosefWN, line 71: Should we use the `key` here (and for similar comparisons), or will these comparisons work even if a prefix is set?

JKRhb:
> LGTM :)

Yay :) I now also cleaned up the commit history so this PR should be ready to be merged :)

shamblett:
Yep I'll have a look, I was also thinking of 4.1.0 for this.

shamblett:
Merged, package re released at version 4.1.0

Issue #78, opened 2022-04-19
Specifying client timeout


JosefWN:
The `client.timeout` overrides the timeout implicitly specified by the config, should we allow this/do we need this? Initial discussion: https://github.com/shamblett/coap/pull/76#discussion_r852652408

I propose just dropping the `client.timeout` and letting the config dictate timeouts. Easier to manage in code and less confusing to the users of the library.

shamblett:
Package re released at version 4.0.0

PR #87, opened 2022-05-03
Remove timeout, code cleanup

JosefWN:
Proposal for https://github.com/shamblett/coap/issues/78

- Now only relying on specified timeouts in `CoapConfig`
- Requests return null response on timeout
- The utils actually made it more complicated to manipulate lists, so I removed those utils and shortened the code a bit (I find the standard library easier to read than custom utils, too)
- Changed some nullables to non-nullables
- Removed some unused code

You can still ignore an ongoing request (equivalent to the previous timeout):

```dart
var respFuture = client.get('test');
respFuture.then((CoapResponse? resp) => ...);
// ...
respFuture.ignore();
```

@JKRhb @shamblett thoughts?


JKRhb (approved):
Thank you, @JosefWN, I think this is another great improvement :)

lib/src/util/coap_util.dart
```diff
@@ -7,88 +7,8 @@
...
-      if (item != source.last) {
-        sb.write(',');
-      }
-    }
-    return sb.toString();
-  }
-
   /// Stringify options in a message.
   static String optionsToString(CoapMessage msg) {
     final sb = StringBuffer();
```
JKRhb, line 14: On second glance, maybe this method could also be moved to the `CoapMessage` class? If I see it correctly, that is the only place where the method is being used.

lib/src/util/coap_util.dart
```diff
@@ -7,88 +7,8 @@
...
-
-  /// The timer callback
-  void _timerCallback() {
-    _running = false;
-    _completer.complete();
-  }
-}
-
 /// Utility methods
 class CoapUtil {
```
JKRhb, line 11: I think in this class, you could also remove the `regIP` member and use `InternetAddress.tryParse(host) == null` to perform the IP address check.

JKRhb: The remaining `lookupHost` method could then be moved to the `CoapClient` class.

lib/src/util/coap_util.dart
JosefWN: Nice catch, fixed: https://github.com/shamblett/coap/pull/87/commits/7136515b45582fce34bce0c8746722ae2ae23f81

shamblett:
OK guys happy with this, I'll merge rt if you're ready.

JosefWN:
Added some minor tweaks for cancelled requests and pings (since token is empty I'm matching on id instead): https://github.com/shamblett/coap/pull/87/commits/280d636b7404c697696d1120407affc13eb81f84.

Ready for merge! :)

PR #88, opened 2022-05-15
refactor!: clean up `CoapMessage` class

JKRhb:
I noticed some more clean up potential in the `CoapMessage` class, due to the fact that you do not need to return the current object for chaining method calls because of Dart's cascade notation. I also managed to get rid of a couple of null check operators, making the package a bit more reliable.

Technically, this PR introduces small but breaking API changes, so I guess it should be deferred to a point when more critical API changes have accumulated for a new major release.

CC @JosefWN 

Depends on #89.


lib/src/coap_message.dart
```diff
@@ -148,8 +146,10 @@ class CoapMessage {
   typed.Uint8Buffer? get token => _token;
 
   /// As a string
-  String get tokenString =>
-      _token != null ? CoapByteArrayUtil.toHexString(_token!) : '';
+  String get tokenString {
+    final token = _token;
```
JosefWN, line 150: What is the purpose of this variable?

lib/src/coap_message.dart
```diff
@@ -292,9 +292,10 @@ class CoapMessage {
 
   /// The payload of this CoAP message in string representation.
   String? get payloadString {
-    if (payload != null && payload!.isNotEmpty) {
+    final payload = this.payload;
```
JosefWN, line 295: Also here, not entirely clear on this one.

lib/src/coap_message.dart
JKRhb: It is simply introduced for type promotion purposes, making it possible to get rid of the null check operator. This is not a must, as `_token` is obviously not `null` when the operator is used, but it is one potential source for bugs less in the future :)

lib/src/coap_message.dart
JKRhb: Same reason as above, the linter does not complain anymore when the null check is removed a couple of lines below.

lib/src/coap_message.dart
JosefWN: Ah, gotcha.

JosefWN:
Yeah I actually thought about doing the same thing, good thinking 👍

Chaining isn't that critical in Dart anyway since you can chain either way with `..`. LGTM.

JKRhb:
I now rebased this PR against #89, so that the two PRs can be merged subsequently.

Issue #80, opened 2022-04-19
Use enhanced enums


JosefWN:
When Dart 2.17 is released, enhanced enums will simplify our code a bit: https://github.com/dart-lang/sdk/issues/47849

Examples:
- [x] Block option SZX could be an int enum
- [x] `CoapCode` could be broken down into int enums
- [x] `CoapMediaType` could be int enum
- [x] `CoapMessageType` could be int enum
- [ ] `Request.observe` could be int enum?
- [x] `CoapDefinedAddress` could be string enum?

JKRhb:
Dart 2.17 has just been released by the way :) https://medium.com/dartlang/dart-2-17-b216bfc80c5d

shamblett:
Yes, got it, FFI updates and experimental RISCV support are good news for me with my IOT hat on.

JKRhb:
I just noticed that the option type numbers might also be an interesting place where the enhanced enums could be used. I'll try to come up with a PR for that soon.

JKRhb:
I started working on refactoring the `CoapCode`s using enhanced enums by the way :)

JosefWN:
A couple of minor things I noticed, which will likely resolve themselves with the re-write, but mentioning just in case. Not all codes are represented in `codeToString` (at least `continues` is missing), and the fallback doesn't show the code:

```dart
    if (isValid(code)) {
      if (isRequest(code)) {
        return 'Unknown Request [code {0}]'; // `$code` rather than `{0}`?
      // ...
```

JKRhb:
> 
> - Block option SZX could be an int enum
> - CoapCode could be broken down into int enums
> - CoapMediaType could be int enum
> - CoapMessageType could be int enum
> - Request.observe could be int enum?
> - CoapDefinedAddress could be string enum?

I think from this list, we could mark `CoapCode`, `CoapMediaType`, and `CoapMessageType` as resolved now :) Should we maybe turn this issue into a task list for tracking the remaining candidates for enhanced enums?

JosefWN:
Done! :)

shamblett:
Package re released at version 5.0.0.

shamblett:
Yes no.problem  a bit overly aggressive with my tidy up.

JKRhb:
> Block option SZX could be an int enum

This is now also addressed by #122 and can be checked once the PR has been merged :)

JosefWN:
LGTM!

PR #89, opened 2022-05-16
Use enhanced enums for option types, handle unknown option numbers

JKRhb:
As mentioned in #80, this PR lets the library use enhanced enums for option types, making the respective code a bit more concise.

While I was at it, I added some additional logic for handling options that are unknown to the client in accordance to [RFC 7252, section 4.2](https://datatracker.ietf.org/doc/html/rfc7252#section-4.2). Such options will now be ignored (in the case of elective options) or will cause responses to be rejected (in the case of critical options).

I am looking forward to your feedback :)

// CC @JosefWN


lib/src/coap_message.dart
```diff
@@ -122,25 +125,18 @@ class CoapMessage {
...
-    for (final option in options) {
-      if (option.type == optionType) {
-        return option;
-      }
-    }
-    return null;
+  CoapOption? getFirstOption(OptionType optionType) {
+    return getOptions(optionType)
+        ?.toList()
+        .firstWhere((element) => element.type == optionType);
```
JosefWN, line 131: Maybe we have to use `firstWhereOrNull` here, `firstWhere` will throw an exception if it doesn't find a match:

> If no element satisfies test, the result of invoking the orElse function is returned. If orElse is omitted, it defaults to throwing a [StateError](https://api.dart.dev/stable/2.0.0/dart-core/StateError-class.html).

https://api.dart.dev/stable/2.0.0/dart-core/Iterable/firstWhere.html

JosefWN:
Great work, looks good! Very nice with added type safety and improved readability!

lib/src/coap_message.dart
JKRhb: Good catch, fixed in 7648235c38be2a02a2bda7c35c10a21d05afe761 :)

PR #90, opened 2022-05-21
chore!: use import/export instead of part/part of

JKRhb:
This is another large cleanup PR, replacing the `part` and `part of` statements with `import` and `export` statements. This greatly reduces the API surface of the library, exposing less of its internals, while making it more transparent which parts of the library depend on each other or external libraries.


JKRhb:
CC @JosefWN :)

JosefWN:
I agree that making the per-file dependencies more explicit is for the better. Haven't tried the examples to see that they work, but I am in favor of the refactoring. LGTM 👍

JKRhb:
> I agree that making the per-file dependencies more explicit is for the better. Haven't tried the examples to see that they work, but I am in favor of the refactoring. LGTM 👍

Thank you for your feedback :) I now also added the pre-defined configs to the list of exports. We might have to take a closer look if there are any missing exports before merging/releasing a new version. At least all the examples should still work, though :)

JKRhb:
@JosefWN @shamblett  Do you think we could merge #90, #91, and #92? Or do you see anything that should be improved?

JosefWN:
Looks fine to me!

shamblett:
HI @JosefWN I'll have a look at merging the PR's you mention above over the next few days

shamblett:
OK, no probs, I'm wary of merging 91 and 92 at the mo I don't want introduce too much change until we've stabilized a bit, unless you think it wouldn't matter for these warnings.

shamblett:
OK, #91 and #92 now merged.

PR #92, opened 2022-05-21
feat!: let failed requests throw Exceptions instead of returning null

JKRhb:
This PR is supposed to be merged after #90 and #91.

Working with the new API, I noticed that handling the possible `null` value that is returned when a request fails is not as convenient as when the request simply throws an Exception. This PR therefore proposes yet another API change by introducing two new Exceptions for cancelled and timed out requests, allowing these two cases of errors to be handled differently.

This behavior should also be more similar to the `http` package's client, which also does not return `null` but throws an exception on failed requests (see, for instance, [here](https://github.com/dart-lang/http/blob/87d43798ae0fe7264ecda9d0dbc3e491a2abab65/lib/src/io_client.dart#L26)).

CC @JosefWN

### DIFF
--- a/example/cancel_request.dart
+++ b/example/cancel_request.dart
@@ -34,14 +34,14 @@ FutureOr<void> main(List<String> args) async {
     ignoreThisFuture.ignore();
 
     final resp = await helloRespFuture;
-    print('/hello response: ${resp?.payloadString}');
+    print('/hello response: ${resp.payloadString}');
 
     if (cancelThisReq.retransmits > 0) {
       print('Expected 0 retransmits!');
     }
-
-    client.close();
   } catch (e) {
     print('CoAP encountered an exception: $e');
   }
+
+  client.close();
 }
--- a/example/delete_resource.dart
+++ b/example/delete_resource.dart
@@ -24,10 +24,10 @@ FutureOr<void> main(List<String> args) async {
     print('Sending delete /test to ${uri.host}');
     final response = await client.delete('test');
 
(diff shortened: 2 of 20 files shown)


example/cancel_request.dart
```diff
@@ -34,14 +34,14 @@ FutureOr<void> main(List<String> args) async {
...
     if (cancelThisReq.retransmits > 0) {
       print('Expected 0 retransmits!');
     }
-
-    client.close();
   } catch (e) {
     print('CoAP encountered an exception: $e');
   }
+
+  client.close();
```
JKRhb, line 46: I moved the close method below the try-catch-blocks in the examples, as otherwise the client is not closed when an exception is thrown (and caught).

example/delete_resource.dart
```diff
@@ -24,10 +24,10 @@ FutureOr<void> main(List<String> args) async {
     print('Sending delete /test to ${uri.host}');
     final response = await client.delete('test');
 
-    print('/test response status: ${response?.statusCodeString}');
-
-    client.close();
+    print('/test response status: ${response.statusCodeString}');
```
JKRhb, line 27: Throwing an exception makes it possible to get rid of the `?`s in the examples

JosefWN:
Yeah, I was a bit on the fence with whether to make responses nullable or throw an exception. I agree this makes more sense, it also makes it possible to differentiate between a timeout and a cancellation when using the shorthands like `client.get('...')`, where the request object is not available for inspection.

PR #93, opened 2022-05-22
feat!: set URI for each request instead of for the client

JKRhb:
This PR depends on #92 and is supposed to be merged after it.

Here, I wanted to make another proposal for a major overhaul of the API, moving the indication of the request URI to the request itself, making it possible to reuse one client for multiple requests to different servers. This makes the handling of the client more intuitive in my opinion, should enhance the performance, and also aligns the library a bit more with the `http` library.

### DIFF
--- a/example/cancel_request.dart
+++ b/example/cancel_request.dart
@@ -13,28 +13,29 @@ import 'config/coap_config.dart';
 
 FutureOr<void> main(List<String> args) async {
   final conf = CoapConfig();
-  final uri = Uri(scheme: 'coap', host: 'coap.me', port: conf.defaultPort);
-  final client = CoapClient(uri, conf);
+  final client = CoapClient(conf);
 
-  final cancelThisReq = CoapRequest.newGet();
-  cancelThisReq.addUriPath('doesNotExist');
+  final cancelUri = Uri.parse("coap://coap.me/doesNotExist");
+  final helloUri = Uri.parse("coap://coap.me/hello");
+
+  final cancelThisReq = CoapRequest.newGet(cancelUri);
 
   try {
     // Ensure this request is not also cancelled
-    print('Sending async get /hello to ${uri.host}');
-    final helloRespFuture = client.get('hello');
+    print('Sending async get ${helloUri.path} to ${helloUri.host}');
+    final helloRespFuture = client.get(helloUri);
 
-    print('Sending async get /doesNotExist to ${uri.host}');
+    print('Sending async get ${cancelUri.path} to ${cancelUri.host}');
     final ignoreThisFuture = client.send(cancelThisReq);
 
-    print('Cancelling get /doesNotExist retries');
+    print('Cancelling get ${cancelUri.path} retries');
     client.cancel(cancelThisReq);
 
-    print('Ignoring /doesNotExist response future');
+    print('Ignoring ${cancelUri.path} response future');
     ignoreThisFuture.ignore();
 
     final resp = await helloRespFuture;
-    print('/hello response: ${resp.payloadString}');
+    print('${helloUri.path} response: ${resp.payloadString}');
 
     if (cancelThisReq.retransmits > 0) {
       print('Expected 0 retransmits!');
--- a/example/delete_resource.dart
+++ b/example/delete_resource.dart
@@ -13,16 +13,12 @@ import 'config/coap_config.dart';
 
 FutureOr<void> main(List<String> args) async {
   final conf = CoapConfig();
-  final uri = Uri(
-    scheme: 'coap',
-    host: 'californium.eclipseprojects.io',
-    port: conf.defaultPort,
-  );
-  final client = CoapClient(uri, conf);
+  final uri = Uri.parse("coap://californium.eclipseprojects.io/test");
+  final client = CoapClient(conf);
 
   try {
-    print('Sending delete /test to ${uri.host}');
-    final response = await client.delete('test');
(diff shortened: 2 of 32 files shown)


JosefWN:
There are a couple of reasons I think this might be a bad idea:
- CoAP is more dependent on aligning server and client configurations than HTTP (retries, timeouts, message/block size limits). Different connectivity, payloads etc. might mean different configs for different URI's. I think it's better to make sure the users configure this explicitly rather than assume one config will work everywhere, giving the user no means of specifying more than one config.
- CoAP is a more niche protocol than HTTP, one client doesn't typically interact with a whole host of servers, only one or a few, in which case creating multiple clients is not a big problem for the library user.

I see a limited use-case for this, and I think it could be so limited that it doesn't justify the added complexity of the code (as opposed to just creating another client). For every server you talk to, I think you _should_ consider the config, and if you pass the config along with every request as well you are essentially creating new clients for all requests.

I also think this PR might not be working right now, it's creating the endpoint in the client with the URI from the _first_ request, the endpoint doesn't change for subsequent requests from what I could see. You have to support multiple endpoints in the client, and only cache them for some duration of idle time to prevent memory leaks in the client if you talk to a lot of endpoints.

I guess what I'm trying to say is if we repeatedly add things that seemingly have no urgent use-case but may or may not come in handy one day, the code base will swell quite a bit over time, and we will be maintaining a lot of unused code or unnecessary complexity.

JKRhb:
> I also think this PR might not be working right now, it's creating the endpoint in the client with the URI from the _first_ request, the endpoint doesn't change for subsequent requests from what I could see. You have to support multiple endpoints in the client, and only cache them for some duration of idle time to prevent memory leaks in the client if you talk to a lot of endpoints.

Ohh, you are right, that is actually something I didn't take into account (the examples still work, which misled me a bit). This is definitely something that would need to be fixed/reworked.

> * CoAP is a more niche protocol than HTTP, one client doesn't typically interact with a whole host of servers, only one or a few, in which case creating multiple clients is not a big problem for the library user.

Depends on the use-case, I would say. You could also have a lot of sensor nodes, for example, which you discovered using a CoRE RD. In this case, re-using a client might be useful and would probably also have performance benefits.

> * CoAP is more dependent on aligning server and client configurations than HTTP (retries, timeouts, message/block size limits). Different connectivity, payloads etc. might mean different configs for different URI's. I think it's better to make sure the users configure this explicitly rather than assume one config will work everywhere, giving the user no means of specifying more than one config.

This is true, but the change would not prevent users from still creating a separate client for each endpoint. Also, we could add the possibility of passing a custom config per request or making them more configurable again. I guess I would also tend to assume that for most use cases the default config is fine or that users only want to override one specific parameter. To do so, you currently need to create your own config class, though, overriding the defaults. Using constructor parameters, you can make the config values configurable at runtime, but I would say that there is still some room for API improvement, even if we stick to the current approach.

JKRhb:
cc6257c1f09d05ebebf28cd9c2c47bbb70254d9f fixes the multiple endpoints issue, caching them by URI host, port, and scheme. This does not allow proper socket reuse yet, but at least the solution is not completely broken now ;) I also added a testable example in 46b07db

JosefWN:
As I started writing an answer I realized that we will probably both get drawn into speculation a bit when we argue for what we believe in. Perhaps it's easier if we come up with an example use-case that demonstrates the need (best of all if it's a real-world example and not a synthetic one, but even a synthetic one is better than having none).

You could show how you would solve that with the API proposed in this PR and I could come with a counter-example of how I would solve it with the existing API. That way we can also examine things a bit further (such as benchmarks).

Assuming that it is something we can already do fairly easily and without performance issues with the current API, I'm leaning toward just showing the users how to achieve this with an example. If we do find problems, my first approach would be to build an abstraction, like a `ClientGroup` to solve that particular problem, rather than making the client more complex for everyone. The client is more or less an endpoint with a config anyway, having a list of clients is pretty much equivalent to have a client with lists of configs and endpoints. The biggest benefit is that you only opt-in to this complexity when you need it. With your PR the complexity would be there for everyone, and with it would come the cost/risk associated with that complexity.

> This is true, but the change would not prevent users from still creating a separate client for each endpoint. Also, we could add the possibility of passing a custom config per request or making them more configurable again.

I know, but the mindset is important I think. One speculative feature is fine, say it adds 500 lines of code. 10-100 speculative features over a few years could increase the code base by tens of thousands of lines of code. I think it's quite a common struggle projects face, for example my refactoring in this repo reduced the code base by several thousand lines even excluding the example cleanup.

We _can_ add a lot of features, but for every feature we consider, we need to weigh the cost over time of maintaining that feature, against the value it will provide to the users. The users won't mind configurability, but I'm saying that unless we are cautious and only implement what we really need, _we_ will mind that configurability. The biggest cost is not typically in the initial implementation of a feature, but in the maintenance of that feature over time.

Even if it's slight, I also think this particular feature carries a performance penalty for everyone who is not using multiple URI's (a more complex hot path), so there is a cost for all users of having a feature only a few or possibly no users will utilize.

JKRhb:
@JosefWN Thank you for your thoughts on this issue! I think I will convert the PR to a draft for now, as the current API works and discussing the concrete changes is a bit difficult due to the included changes from other PRs. I also realized that this PR should also include some changes to the internal endpoint API to actually have the intended effect, which would require a bit more work. Therefore, I suggest putting this one on hold until the changes have a more solid foundation :)

JKRhb:
Superseded by #160.

PR #97, opened 2022-06-17
Upgrade dependencies

JosefWN:
Was unsure about DTLS dependencies so left them for now, figured @JKRhb would know better for those.


JKRhb:
Thank you, @JosefWN :) I opened #98 for `dart_tinydtls` – I actually thought before there were breaking changes, but it still works without adjusting the implementation here :)

PR #101, opened 2022-06-21
fix: change identifier of OptionType name field

JKRhb:
In the current Dart version, using a `name` field in an enhanced enum (as we currently do in the `OptionType` enum) causes a segmentation fault (c.f., https://github.com/dart-lang/sdk/issues/49296). Although this will be fixed in the next version, this PR simply changes the identifier of the name in question as a workaround to avoid segfaults.

### DIFF
--- a/lib/src/coap_option.dart
+++ b/lib/src/coap_option.dart
@@ -95,7 +95,7 @@ class CoapOption {
   }
 
   /// Gets the name of the option that corresponds to its type.
-  String get name => _type.name;
+  String get name => _type.optionName;
 
   /// Gets the value's length in bytes of the option.
   int get length => _buffer.lengthInBytes;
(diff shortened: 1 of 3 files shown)


JosefWN:
Since it's already merged maybe we could wait for the fix to be released? Just using `name` looks a little prettier :)

JKRhb:
> Since it's already merged maybe we could wait for the fix to be released? Just using `name` looks a little prettier :)

Yeah, that's true :) However, it seems as if the bug is still present in Dart 2.17.5 (see [here](https://github.com/namib-project/dart_wot/runs/7135598529?check_suite_focus=true)) :/ I hope it will be part of a 2.17.6 release

JKRhb:
@JosefWN Hmm, in the issue, the contributor of the fix [wrote](https://github.com/dart-lang/sdk/issues/49296#issuecomment-1172503136) that it will only be included in Dart 2.18, which seems to be scheduled for August :/ So I think it would be good if we could merge this PR already to be able to continue with adding new features/refactorings.

Issue #100, opened 2022-06-19
Multicast requests using blockwise currently not working


JKRhb:
In the current version of the library, using multicast with block-wise transfer currently does not work anymore, as the client checks if the response's token matches the one of the original request.

https://github.com/shamblett/coap/blob/3f03f45e66ca4102ce729a95a0477fab4acd1292/lib/src/coap_client.dart#L474

However, because the client switches to unicast when using block-wise transfer with a server, this check always returns `false` because a new token is assigned once the switch is made. I will try to come up with a fix for this problem in the next couple of days.

Maybe the check could simply return `true` if multicast is being used, like so?

```dart
.where((final e) => req.multicast || e.resp.token!.equals(req.token!))
```

Or could we add a field for an "original token" somewhere, which could be used for matching instead?

JosefWN:
Ah, ouch. The token for the "overarching" request is set in the exchange when `fireRespond` is called: https://github.com/shamblett/coap/blob/4.2.1/lib/src/net/coap_exchange.dart#L154

For blockwise this is the last token in the blockwise exchange.

I can test a bit too. Just for reference, server that could be used for testing (can be modified to return blockwise): https://github.com/plgd-dev/go-coap/tree/master/examples/mcast/server

Client:
```dart
FutureOr<void> main(final List<String> args) async {
  final conf = CoapConfig();
  final uri = Uri(
    scheme: 'coap',
    host: CoapDefinedAddress.allCOAPNodesIPV4,
    port: conf.defaultPort,
  );
  final client = CoapClient(uri, conf);

  try {
    final req = CoapRequest.newGet()..uriPath = '/oic/res';
    await client.send(
      req,
      onMulticastResponse: CoapMulticastResponseHandler(print),
    );
  } on Exception catch (e) {
    print('CoAP encountered an exception: $e');
  }

  client.close();
}
```

JosefWN:
Ah, my bad, I was printing all of the events and must have mistaken `CoapRespondingEvent` for `CoapRespondEvent`. Will dig a bit more, it could have to do with how the exchange is duplicated.

JosefWN:
Yeah, tried and it works in `v3.5.0`, so it was my refactor.

An attempt at a fix with a multicast token like you were thinking of: https://github.com/shamblett/coap/pull/102

JosefWN:
We can keep this open until this is fixed:
[...]

PR #102, opened 2022-06-23
Fix for multicast block2 responses

JosefWN:
Attempt at fixing: https://github.com/shamblett/coap/issues/100

### DIFF
--- a/lib/src/coap_client.dart
+++ b/lib/src/coap_client.dart
@@ -331,7 +331,11 @@ class CoapClient {
       }
       _eventBus
           .on<CoapRespondEvent>()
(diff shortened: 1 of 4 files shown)


JKRhb:
Awesome, thank you :) Looks good already, I will try it out as soon as possible :)

JKRhb:
Thank you, @JosefWN, block-wise multicast requests now work again :) The only thing I noticed is that the client seems to hang if you use multicast even if you explicitly close it. Unfortunately, I haven't been able to find the point where to fix the issue yet.

JosefWN:
Odd, I can take a look, but open a separate PR for that!

Issue #103, opened 2022-06-25
Method not allowed with post request


Dodecanom:
Hello,

I’m trying to connect to my Ikea lights with Flutter. My issue is that when I try to do a post request the response is `Code: 4.05 Method Not Allowed` which is weird because a post request should be allowed.

When I do the exact same request from command line there is no issue (with the SECURITY_CODE and PRE_SHARED_KEY replaced of course) :

Request: `coap-client-openssl -m post -u "Client_identity" -k "SECURITY_CODE" -e '{"9090":"IDENTITY"}' "coaps://192.168.1.41:5684/15011/9063"`
Response: `{"9091":"PRE_SHARED_KEY","9029":"1.19.0026"}`

Could you have a look to my code and tell me what’s wrong? I would be very grateful if you could help me to get this working, I really want to play with this!

If you are curious the request I try to do is the one described here: https://github.com/home-assistant-libs/pytradfri/issues/90

Here is a minimum example followed by my pubspec

```dart
import 'package:flutter/material.dart';
import 'package:coap/coap.dart';

void main() {
  runApp(const MyApp());
}

class MyApp extends StatelessWidget {
  const MyApp({Key? key}) : super(key: key);

  @override
  Widget build(BuildContext context) {
    return MaterialApp(
        title: 'Test lights',
        theme: ThemeData(
          primarySwatch: Colors.blue,
        ),
        home: const TestWidget());
  }
}

class TestWidget extends StatefulWidget {
  const TestWidget({
    Key? key,
  }) : super(key: key);

  @override
  State<TestWidget> createState() => _TestWidgetState();
}

class _TestWidgetState extends State<TestWidget> {
  @override
  Widget build(BuildContext context) {
    return SafeArea(
      child: Scaffold(
        body: TextButton(
          child: const Text('Test button'),
          onPressed: () async {
            final conf = DtlsConfig();

            var credentials = PskCredentials(
              identity: 'Client_identity',
              preSharedKey: 'SECURITY_CODE',
            );

            PskCredentials pskCredentialsCallback(final String indentity) => credentials;

            final uri = Uri(
              scheme: 'coaps',
              host: '192.168.1.41',
              port: conf.defaultSecurePort,
            );

            final client = CoapClient(
              uri,
              conf,
              pskCredentialsCallback: pskCredentialsCallback,
            );

            try {
              client.events.on<Object>().listen(print);

              final response = await client.post(
                '15011/9063',
                payload: '{"9090":"IDENTITY"}',
[...]

PR #104, opened 2022-06-26
Fix issue 103, improve validation

JosefWN:
- `addUriPath` appends a single path to the full path, `uriPath` sets the path as a whole. Since `addUriPath` is a bit less idiomatic to use when setting the full path I have removed that use from the examples (and by mistake in the `CoapClient` as well, the root cause for the bug in #103).
- Setting the `uri` (internal setter) would because of my re-write override the `uriPath` with that of the URI (blank), removed this since it's the base URI that's actually set (host/port).
- Removed some nullables where the function would throw an error if the value was null.
- Made properties symmetric, i.e. `locationPath` setter and getter both return the same thing (almost, sanitized input), rather than setting `locationPath` and getting the same value from `locationPathsString`.
- Removed _nearly_ duplicate properties (`uriPath` vs `uriPathsString`).
- Improved input validation (for example uri path length check only took place on `addUriPath`, but not when setting `uriPath`). Added tests to cover edge cases.


test/coap_message_api_test.dart
```diff
@@ -242,77 +242,109 @@ void main() {
...
     expect(message.uriPaths.length, 4);
-    expect(message.uriPathsString, 'a/uri/path/longer');
-    expect(() => message.addUriPath(null), throwsArgumentError);
+    expect(message.uriPath, 'a/uri/path/longer');
+    expect(
+      () => message.addUriPath('multiple/not/allowed'),
+      throwsArgumentError,
+    );
+    expect(
+      () => message.addLocationPath('no-double-slash//'),
```
JosefWN, line 258: Technically I think an URI path option can be empty, so `a//b/c` might be a valid URI. Not sure how to feel about that, the reason I added this check is I didn't want people adding multiple paths using `addLocationPath` (since it's intended for adding a single path option or path segment).

shamblett:
I wasn't sure if this was ready, so I merged the others, this has now resulted in a conflict.

JosefWN:
Rebased, added a couple more test cases and re-ran all examples except the DTLS ones and they looked OK.

I'm fine with merging this if you guys are.

PR #106, opened 2022-06-26
refactor!: use enhanced enum for ContentMediaType

JKRhb:
This reworks the Content-Format registry as an enhanced enum and adds the missing values from the IANA CoAP [Content-Formats registry](https://www.iana.org/assignments/core-parameters/core-parameters.xhtml#content-formats), also supporting those with parameters and encoding values.

CC @JosefWN

### DIFF
--- a/lib/src/coap_client.dart
+++ b/lib/src/coap_client.dart
@@ -129,7 +129,7 @@ class CoapClient {
   /// Sends a GET request.
   Future<CoapResponse> get(
     final String path, {
-    final int accept = CoapMediaType.textPlain,
+    final CoapMediaType accept = CoapMediaType.textPlain,
     final int type = CoapMessageType.con,
     final List<CoapOption>? options,
     final bool earlyBlock2Negotiation = false,
@@ -153,8 +153,8 @@ class CoapClient {
   Future<CoapResponse> post(
     final String path, {
     required final String payload,
-    final int format = CoapMediaType.textPlain,
-    final int accept = CoapMediaType.textPlain,
+    final CoapMediaType format = CoapMediaType.textPlain,
+    final CoapMediaType accept = CoapMediaType.textPlain,
     final int type = CoapMessageType.con,
     final List<CoapOption>? options,
     final bool earlyBlock2Negotiation = false,
@@ -178,8 +178,8 @@ class CoapClient {
   Future<CoapResponse> postBytes(
     final String path, {
     required final Uint8Buffer payload,
-    final int format = CoapMediaType.textPlain,
-    final int accept = CoapMediaType.textPlain,
+    final CoapMediaType format = CoapMediaType.textPlain,
+    final CoapMediaType accept = CoapMediaType.textPlain,
     final int type = CoapMessageType.con,
     final List<CoapOption>? options,
     final bool earlyBlock2Negotiation = false,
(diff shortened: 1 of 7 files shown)


JKRhb:
Note that I removed the `fileExtension` field here as Content-Formats can have multiple file extensions and keeping track of all possible extensions seemed a bit tedious to me, while not being sure if it has that much value for the users. If you think we should keep the extension, we could add them back in, though (maybe in that case as a list, though).

test/coap_base_class_test.dart
```diff
@@ -57,14 +25,12 @@ void main() {
     });
 
     test('Parse wild card', () {
-      var res = CoapMediaType.parseWildcard(null);
-      expect(res, isNull);
-
-      res = CoapMediaType.parseWildcard('xml*');
-      expect(res, <int>[
+      final res = CoapMediaType.parseWildcard('xml*');
```
JosefWN, line 28: Have only had a super brief look, but couldn't we have a `CoapMediaType.startsWith` instead, which is using the string method `startsWith`? It would perform better than a regex, and the code would be a bit simpler.

JosefWN:
LGTM, just a minor remark.

test/coap_base_class_test.dart
JKRhb: Hmm, I think a `startsWith` method would not really catch the same semantics, as the current version of this method seems to also (or only?) find those Content-Formats that *end* with `xml`. Maybe we could also remove the method completely, though, since users could also use something like `CoapMediaType.values.where((value) => value.mimeType.endsWith('xml'))`.

JosefWN: Oh, I thought `xml*` would match anything starting with `xml` and `*xml` would do the opposite.

Yeah, in general I think this method (and several others like it) may have made sense way back in .NET before linq/enumerables, but now in Dart it feels a bit superflous to re-implement this type of logic, `.where(...)` is providing a lot more flexibility with a small increase in verbosity. Using the Dart SDK also makes it clearer to the people using the lib what's going on.

JKRhb: > Oh, I thought `xml*` would match anything starting with `xml` and `*xml` would do the opposite.

Hmm, yeah, you are right, that should be the case. I just got the impression that this is the current behavior, as all returned formats end with `xml`. Maybe it currently matches all Content-Formats that contain `xml`? The implementation does not seem that intuitive in general.

> Yeah, in general I think this method (and several others like it) may have made sense way back in .NET before linq/enumerables, but now in Dart it feels a bit superflous to re-implement this type of logic, .where(...) is providing a lot more flexibility with a small increase in verbosity. Using the Dart SDK also makes it clearer to the people using the lib what's going on.

That's true! Then I will clean up the file a bit more and remove the method :)

lib/src/coap_client.dart
```diff
@@ -129,7 +129,7 @@ class CoapClient {
   /// Sends a GET request.
   Future<CoapResponse> get(
     final String path, {
-    final int accept = CoapMediaType.textPlain,
+    final CoapMediaType accept = CoapMediaType.textPlain,
```
JKRhb, line 132: One general question I was wondering about: Using the enum value here limits users to the pre-defined values. While CoAP implementations should ignore all unknown Content-Formats, on second thought, maybe this approach is a bit too strict, and we should keep numeric values both as inputs and outputs, using something like this instead:

```suggestion
    final int accept = CoapMediaType.textPlain.numericValue,
```

In CoAP messages, we could add a field for both the numeric value and the parsed `CoapMediaType` (which would return `null` if the value is unknown`).

What do you think, @JosefWN?

JosefWN: I think it's fine to constrain the input, it's worth the added type safety, the risk of making mistakes is likely greater than the probability that someone encounters a type that we don't support (and if they do, they could always open a ticket).

JKRhb: Sounds reasonable :) Thank you for the feedback!

test/coap_base_class_test.dart
JKRhb: Method is now removed :)

Issue #107, opened 2022-07-06
Since version 4.2.0 observe non confirmable not working properly


vincent-iQontrol:
I use a non confirmable observer from the example. Version 4.0.0 and 4.1.0 is working correctly, all events are coming in. In 4.2.0 and above a lot of events are missing after ~ 13 events.

final reqObsNon = CoapRequest(CoapCode.get, confirmable: false)..addUriPath('Json');

JosefWN:
Confirmed using this snippet (works in 4.1.0 but not in 4.2.0):
```dart
import 'dart:async';
import 'package:coap/coap.dart';
import 'config/coap_config.dart';

FutureOr main() async {
  final conf = CoapConfig();
  final uri = Uri(
    scheme: 'coap',
    host: 'californium.eclipseprojects.io',
    port: conf.defaultPort,
  );
  final client = CoapClient(uri, conf);

  try {
    final reqObsNon = CoapRequest(CoapCode.get, confirmable: false);
    reqObsNon.addUriPath('obs-non');

    print('Observing /obs-non on ${uri.host}');
    final obsNon = await client.observe(reqObsNon);
    obsNon.stream.listen((CoapRespondEvent e) {
      print('/obs-non response: ${e.resp.payloadString}');
    });

    await Future.delayed(Duration(minutes: 15));
  } catch (e) {
    print('CoAP encountered an exception: $e');
  }

  client.close();
}
```

I will try to dig into it soon.

vincent-iQontrol:
Thanks for investigating. I will use an older version for the time being.

JosefWN:
Made a couple of more test, this bug seems to have been introduced with my stricter linting of the repo...

shamblett:
Yep ok Ill look at the PRs over the weekend ta.

shamblett:
With this and #112 now merged are we ready for another release? There's quite a bitt gone since 4.2.1

JosefWN:
We could merge PR #111 too perhaps? Then I can rebase my other PR on master to reduce its size.

JKRhb:
> With this and #112 now merged are we ready for another release? There's quite a bitt gone since 4.2.1

I also have one or two more upcoming PRs which I will open soon, then we could go for a new release :)

shamblett:
OK PR #111 merged, I'll hold off on the release.

JKRhb:
> 
> I also have one or two more upcoming PRs which I will open soon, then we could go for a new release :)

[...]

PR #109, opened 2022-07-12
Fixes issue #107

JosefWN:
The named constructors were not setting `CoapCode.empty`, converted to factory constructors to re-use the main constructor.

EDIT: Also had to downgrade a couple of dependencies to get things to work with the latest version of Flutter, a bit of an annoyance that they are not up to speed.

### DIFF
--- a/lib/src/coap_empty_message.dart
+++ b/lib/src/coap_empty_message.dart
@@ -18,27 +18,24 @@ class CoapEmptyMessage extends CoapMessage {
 
   /// Create a new acknowledgment for the specified message.
   /// Returns the acknowledgment.
-  CoapEmptyMessage.newACK(final CoapMessage message) {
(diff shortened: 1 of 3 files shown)


JKRhb:
Thanks, @JosefWN, I will try out the changes ASAP :)

> EDIT: Also had to downgrade a couple of dependencies to get things to work with the latest version of Flutter, a bit of an annoyance that they are not up to speed.

I think the downgrade is not actually needed (anymore), you can update the internal dependencies using `flutter update-packages --force-upgrade`, this solved the issue for me.

JosefWN:
Cool, reverted the dependency downgrade commit :)

JKRhb (approved):
Tested out the both the "original" observe example and the one given by you in #107 and both work as expected :)

Issue #110, opened 2022-07-12
Killing network during requests causes uncatchable exception


JosefWN:
We don't seem to check if the socket is bound (`send`'s return value is typically ignored, and it's always -1?), but perhaps this is okay since we have retries anyway. Should we ignore any `SocketExceptions` thrown by `send` as well then?

Once this issue is fixed I need to remember to test if the previously bound socket is still usable, or if we need to re-bind after some downtime. Need to think about error handling for observe as well, on re-bind...

Example, turning my wifi off during this loop (testing what happens if I drop off a shaky network temporarily):
```dart
FutureOr<void> main() async {
  final conf = CoapConfig();
  final uri = Uri(
    scheme: 'coap',
    host: 'californium.eclipseprojects.io',
    port: conf.defaultPort,
  );
  final client = CoapClient(uri, conf);

  try {
    for (var i = 0; i < 10; i++) {
      print('Sending request $i');
      final resp = await client.get('test');
      if (resp.code != CoapCode.content) {
        print('Request failed!');
      }
      await Future<void>.delayed(const Duration(seconds: 5));
    }
  } on Exception catch (e) {
    print('CoAP encountered an exception: $e');
  }

  client.close();
}
```

```
Unhandled exception:
SocketException: Send failed (OS Error: Network is unreachable, errno = 51), address = 0.0.0.0, port = 0
#0      _NativeSocket.send (dart:io-patch/socket_patch.dart:1213:34)
#1      _RawDatagramSocket.send (dart:io-patch/socket_patch.dart:2450:15)
#2      CoapNetworkUDP.send (package:coap/src/network/coap_network_udp.dart:55:16)
#3      CoapEndPoint.sendRequest (package:coap/src/net/coap_endpoint.dart:198:21)
#4      CoapStackBottomLayer.sendRequest (package:coap/src/stack/coap_layer_stack.dart:162:30)
#5      CoapNextLayer.sendRequest (package:coap/src/stack/coap_layer_stack.dart:29:10)
#6      CoapAbstractLayer.sendRequest (package:coap/src/stack/coap_abstract_layer.dart:26:15)
#7      CoapReliabilityLayer.sendRequest (package:coap/src/stack/coap_reliability_layer.dart:118:11)
#8      CoapNextLayer.sendRequest (package:coap/src/stack/coap_layer_stack.dart:29:10)
#9      CoapAbstractLayer.sendRequest (package:coap/src/stack/coap_abstract_layer.dart:26:15)
[...]

PR #111, opened 2022-07-12
Remove executor dependency

JosefWN:
I don't think this is necessary, the executor is also using darts task queue (not isolates), so we could as well use dart's task queue directly? Leaving it as a draft until I've had time to test more, but remove it didn't seem to affect performance at first glance, and the library doesn't seem that popular or well maintained.

The reason I dug into it was because of the uncatchable exceptions I've been seeing in #110 (but this doesn't seem to solve the problem).

```bash
# With executor
dart sync_vs_async.dart
Sending 100 async requests...
100 async requests took 67 ms
Sending 100 sync requests...
100 sync requests took 3965 ms

# Without executor
dart sync_vs_async.dart
Sending 100 async requests...
100 async requests took 65 ms
Sending 100 sync requests...
100 sync requests took 3784 ms
```

It even seems a bit faster, but I haven't run that many benchmarks :)

All examples seem to work (save the observe issues solved by my other PR).


JosefWN:
Speaking of dependencies, we could also consider changing `hex` to `convert` (Dart's official library for conversions to hex and other formats), it's a bit bigger but it looks like it's better maintained and seems like a more optimized/less naive implementation at first glance?

PR #114, opened 2022-07-16
refactor!: rework `CoapCode` and `CoapMessageType` as enhanced enums

JKRhb:
This PR refactors the `CoapCode` and `CoapMessageType` classes as enhanced enums (also see #80), making the code more concise and readable. The changes also make it easier to add more `CoapCode`s in the future. The PR itself adds a few missing `CoapCode`s from the IANA registry, including the new method codes `FETCH`, `PATCH`, and `iPATCH`.

CC @JosefWN

### DIFF
--- a/lib/src/coap_client.dart
+++ b/lib/src/coap_client.dart
@@ -16,6 +16,7 @@ import 'coap_block_option.dart';
 import 'coap_code.dart';
 import 'coap_config.dart';
 import 'coap_constants.dart';
+import 'coap_empty_message.dart';
 import 'coap_link_format.dart';
 import 'coap_media_type.dart';
 import 'coap_message.dart';
@@ -130,18 +131,17 @@ class CoapClient {
   Future<CoapResponse> get(
     final String path, {
     final CoapMediaType accept = CoapMediaType.textPlain,
-    final int type = CoapMessageType.con,
+    final bool confirmable = true,
     final List<CoapOption>? options,
     final bool earlyBlock2Negotiation = false,
     final int maxRetransmit = 0,
     final CoapMulticastResponseHandler? onMulticastResponse,
   }) {
-    final request = CoapRequest.newGet();
+    final request = CoapRequest.newGet(confirmable: confirmable);
     _build(
       request,
       path,
       accept,
-      type,
       options,
       earlyBlock2Negotiation,
       maxRetransmit,
@@ -155,18 +155,18 @@ class CoapClient {
     required final String payload,
     final CoapMediaType format = CoapMediaType.textPlain,
     final CoapMediaType accept = CoapMediaType.textPlain,
-    final int type = CoapMessageType.con,
+    final bool confirmable = true,
     final List<CoapOption>? options,
     final bool earlyBlock2Negotiation = false,
     final int maxRetransmit = 0,
     final CoapMulticastResponseHandler? onMulticastResponse,
   }) {
-    final request = CoapRequest.newPost()..setPayloadMedia(payload, format);
+    final request = CoapRequest.newPost(confirmable: confirmable)
+      ..setPayloadMedia(payload, format);
     _build(
       request,
       path,
       accept,
-      type,
       options,
       earlyBlock2Negotiation,
       maxRetransmit,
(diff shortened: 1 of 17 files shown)


lib/src/coap_client.dart
```diff
@@ -205,20 +205,21 @@ class CoapClient {
     required final String payload,
     final CoapMediaType format = CoapMediaType.textPlain,
     final CoapMediaType accept = CoapMediaType.textPlain,
-    final int type = CoapMessageType.con,
+    final bool confirmable = true,
+    final CoapMessageType type = CoapMessageType.con,
```
JosefWN, line 209: Remove this for `put` since we have the `confirmable` bool now, guessing this is a typo? Could be confusing if `confirmable` is `false` and `type` is `CoapMessageType.con` for example.

lib/src/coap_client.dart
```diff
@@ -305,8 +305,9 @@ class CoapClient {
   /// Discovers remote resources.
   Future<Iterable<CoapWebLink>?> discover({
     final String query = '',
+    final bool confirmable = true,
```
JosefWN, line 308: Not sure if `discover` _can_ be non-confirmable. I'm thinking the response will be a piggybacked ack.

Californium also only seems to support the query argument for the discover method: https://github.com/eclipse/californium/blob/3eecb4c739a40ac1af066a89d863be338fa54e00/californium-core/src/main/java/org/eclipse/californium/core/CoapClient.java#L559

lib/src/codec/decoders/coap_message_decoder.dart
```diff
@@ -63,10 +77,10 @@ abstract class CoapMessageDecoder implements CoapIMessageDecoder {
 
   @override
   CoapResponse? decodeResponse() {
-    if (isResponse) {
-      final response = CoapResponse(code)
-        ..type = type
-        ..id = id;
+    final decodedCode = _decodedCode;
+    final coapMessageType = _decodedType;
+    if (decodedCode != null && coapMessageType != null && isResponse) {
```
JosefWN, line 82: Tiny nitpick, I guess we could exit early here:
```dart
if (!isResponse || decodedCode == null || coapMessageType == null) {
  return null;
}
// ...
```

Won't make a big difference here, but defensive programming and early returns is easier to read, so less error prone I think. One discussion can be found here: https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement

lib/src/codec/decoders/coap_message_decoder.dart
```diff
@@ -75,10 +89,9 @@ abstract class CoapMessageDecoder implements CoapIMessageDecoder {
 
   @override
   CoapEmptyMessage? decodeEmptyMessage() {
-    if (!isResponse && !isRequest) {
-      final message = CoapEmptyMessage(code)
-        ..type = type
-        ..id = id;
+    final coapMessageType = _decodedType;
+    if (coapMessageType != null && isEmpty) {
```
JosefWN, line 93: ... same here

lib/src/coap_client.dart
JKRhb: Thank you, that was actually a typo :)

lib/src/coap_client.dart
JKRhb: Good catch, I think that might have been a copy-and-paste mistake ;)

lib/src/codec/decoders/coap_message_decoder.dart
JKRhb: Oh, yeah, that is much better, thanks, I updated the code accordingly :)

lib/src/codec/decoders/coap_message_decoder.dart
JKRhb: Also updated :)

JKRhb:
I incorporated the suggestions and squashed the commits :)

lib/src/stack/coap_reliability_layer.dart
```diff
@@ -128,24 +123,7 @@ class CoapReliabilityLayer extends CoapAbstractLayer {
     final CoapResponse? response,
   ) {
     final mt = response!.type;
-    if (mt == CoapMessageType.unknown) {
```
JosefWN, line 131: Haven't read all of the code and a little sleepy. When would this happen, if the server sent an invalid response and/or if the response was corrupted in flight? I take it with your change we won't acknowledge that message then, meaning that the server will resend the response? Sounds reasonable

JosefWN:
LGTM!

PR #116, opened 2022-07-16
fix: allow empty URI path segments in requests

JKRhb:
Interacting with other CoAP libraries, we noticed in my project that URI paths seem to not be converted into options correctly. In particular, it seems as if empty URI path segments are supposed to be converted into empty URI-path options. Currently, these empty segments are removed from the URI path completely.

This PR adjusts the corresponding creation method accordingly, which should also make the library's behavior more consistent with [section 6.4 of RFC 7252](https://datatracker.ietf.org/doc/html/rfc7252#section-6.4).

### DIFF
--- a/lib/src/coap_message.dart
+++ b/lib/src/coap_message.dart
@@ -625,8 +625,14 @@ abstract class CoapMessage {
 
   /// Set the location path from a string
   set locationPath(final String fullPath) {
-    final trimmedPath = _trimChar(fullPath, '/');
     clearLocationPath();
+
+    var trimmedPath = fullPath;
+
+    if (fullPath.startsWith('/')) {
+      trimmedPath = fullPath.substring(1);
+    }
+
     trimmedPath.split('/').forEach(addLocationPath);
   }
 
--- a/lib/src/coap_message.dart
+++ b/lib/src/coap_message.dart
@@ -496,8 +496,12 @@ abstract class CoapMessage {
   }
 
   /// URI path
+  // TODO: Apply proper percent-encoding
   String get uriPath =>
-      CoapOption.join(getOptions(OptionType.uriPath), '/') ?? '';
+      getOptions(OptionType.uriPath)
+          ?.map((final e) => e.stringValue.replaceAll('/', '%2F'))
+          .join('/') ??
+      '';
 
   /// Sets a number of Uri path options from a string
   set uriPath(final String fullPath) {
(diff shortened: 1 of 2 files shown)


JosefWN:
Should we do the same for `uriLocationPath`?

JKRhb:
> Should we do the same for `uriLocationPath`?

Good point!

I adjusted the code, I am not entirely sure if this is the correct way to handle this option, though. Let me know if you have any suggestions :)

JosefWN:
We could sneak a peak at Californium :)

https://github.com/eclipse/californium/blob/f4680981c68ec48b2ecb6dc572ec5bd43300a208/californium-core/src/main/java/org/eclipse/californium/core/coap/OptionSet.java#L659

Seems similar to our approach, but if we want to make the code a tiny bit shorter:
```dart
final trimmedPath = fullPath;
if (fullPath.startsWith('/')) {
  trimmedPath = fullPath.substring(1);
}
```

JKRhb:
Good catch, could clean up the `locationPath` setter a lot :)

JosefWN:
> Good catch, could clean up the locationPath setter a lot :)

I think we can do the same for `uriPath`, but it's just nitpicks anyway. The PR LGTM!

JKRhb:
> I think we can do the same for `uriPath`, but it's just nitpicks anyway. 

Cleaned up that method as well now :)

> The PR LGTM!

🎉

lib/src/coap_message.dart
```diff
@@ -503,8 +503,12 @@ class CoapMessage {
   }
 
   /// URI path
+  // TODO: Apply proper percent-encoding
```
JosefWN, line 506: The `convert` library has an encoder/decoder for percent as well (in case we need to consider other characters as well): https://github.com/dart-lang/convert/tree/master/lib/src/percent

JKRhb: Thank you for the link! The problem is that there are very specific instructions for the percent encoding (see https://datatracker.ietf.org/doc/html/rfc7252#section-6.5) and it seems to me that there is no way to define exceptions for which characters are supposed to be encoded :/ (I originally used the `decodeComponent` method, but it also converted sub-delimiters like the `=` sign).

JosefWN:
Great work on RFC compliance in general, and sorry for the sloppy URI/query validation I added in my previous PR. Should have checked the RFC, I've been a bit stressed and busy with other things lately.

PR #118, opened 2022-07-17
fix: don't set default Accept and Content-Format options

JKRhb:
At the moment, both for the Accept and the Content-Format option, a default value of 0 (`text/plain; charset=utf-8`) is set for requests. This is a bit at odds with the CoAP specification, which assumes no default value for these two options and can lead to unintended side effects if a server rejects certain accept values (but would allow requests with no accept option set).

Therefore, this PR proposes removing the default values from the creation methods, giving full control to the library users.

### DIFF
--- a/lib/src/coap_message.dart
+++ b/lib/src/coap_message.dart
@@ -289,7 +289,7 @@ abstract class CoapMessage {
   int get payloadSize => payload?.length ?? 0;
 
   /// The payload of this CoAP message in string representation.
-  String? get payloadString {
+  String get payloadString {
     final payload = this.payload;
     if (payload != null && payload.isNotEmpty) {
       try {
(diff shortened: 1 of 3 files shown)


lib/src/coap_message.dart
```diff
@@ -322,7 +322,7 @@ class CoapMessage {
   }
 
   /// Sets the payload and media type of this CoAP message.
-  void setPayloadMedia(final String? payload, final CoapMediaType mediaType) {
+  void setPayloadMedia(final String? payload, final CoapMediaType? mediaType) {
     if (payload == null) {
```
JosefWN, line 326: To align this with `payloadMediaRaw` perhaps we should go with `final String payload` (non null) and skip the null check.

For both functions we could also make `mediaType` a named argument perhaps? So you can do `setPayloadMedia('abc')` without explicitly setting the `mediaType` to `null`, or `setPayloadMedia('abc', mediaType: ...)` (which is a bit clearer than using an optional argument).

JosefWN:
LGTM, just a minor comment.

lib/src/coap_message.dart
JKRhb: Thanks, good suggestion :) I tried to incorporate it in the latest commit, let me know if you see any potential for improvement :)

JKRhb:
Rebased this PR and #116 to the latest state of `master`, hope that both PRs can be merged again.

PR #119, opened 2022-07-21
Refactor link format implementation, add CoRE RD parameters

JKRhb:
This PR reworks the defined parameters for the CoRE Link Format using enhanced enums and adds the parameters defined in RFC 9176 (CoRE Resource Directory). It also fixes some issues that arose from parameter values which contained, for example, dots, which did not get parsed correctly before. I think the implementation might be able to be reworked a bit more in the future, but for now the bug fix and the missing parameters had the highest priority for now.


JKRhb:
I now added some additional refactoring to this PR (moving all link-format related classes to one sub-package) and added some more exports. It seems to me as if the link-format implementation needs to be revisited/reworked a bit more thoroughly in a follow-up PR. But I think it should already be improvement feature- and structure-wise.

JKRhb:
(Sorry for the multiple rebases, resolved the merge conflict with the master branch.)

PR #122, opened 2022-08-09
Rework option system

JKRhb:
I am currently working on an overhaul of the option implementation. This PR is still WIP but I already wanted to share my progress with you.


shamblett:
Just to hijack this slightly are we now good to go for a 5.0.0 release, there's been a lot put in since 4.2.1

JKRhb:
> Just to hijack this slightly are we now good to go for a 5.0.0 release, there's been a lot put in since 4.2.1

Oh, yeah, sorry, I think we can go forward with a 5.0.0 release :) Although I made some good progress here already, this PR is probably something for the next major release.

JKRhb:
I think this PR would now be ready for an initial review (CC @JosefWN). If you think the PR is close to being mergeable, we could include it in the 5.0.0 release. As mentioned above, we could also include it in a later release.

JKRhb:
Applied a couple of further improvements, making options actually immutable and providing a very basic OSCORE option implementation. I think there is still some more work to be done, so I would definitely postpone this after the 5.0.0 release.

shamblett:
Ok will do Ive targetted the 5.0.0 release for this Friday

JKRhb:
> Ok will do Ive targetted the 5.0.0 release for this Friday

Okay, great :)

lib/src/coap_client.dart
```diff
@@ -131,7 +132,7 @@ class CoapClient {
     final String path, {
     final CoapMediaType? accept,
     final bool confirmable = true,
-    final List<CoapOption>? options,
+    final List<Option<dynamic>>? options,
```
JosefWN, line 135: Could we use `Object` rather than `dynamic`? It's a little more specific.

JKRhb: Had to use `Object?` because of the Empty option format, hope it's still an improvement.

lib/src/option/integer_option.dart
```diff
@@ -0,0 +1,157 @@
...
+  @override
+  final optionFormat = OptionFormat.integer;
+
+  @override
+  final OptionType type;
+  @override
+  final int value;
+
+  /// Note that that integer-based option use network-byteorder, therefore the
+  /// order of the bytes has to be reversed.
```
JosefWN, line 27: Although _most_ architectures use little endian (which reversed becomes big/network endian), some CPU architectures are natively big endian, and some are bi-endian, ARM could run in big endian mode for example.

Just to be sure, we could be specific about the endianness: https://stackoverflow.com/questions/65376533/dart-typeddata-and-big-little-endian-representation

JKRhb: Do you think something like this would work?

```dart
ByteData.sublistView(Uint8List.fromList(byteValue)).getUint16(0, Endian.host);
```

Or do we need two variants, depending on the CPU architecture?

test/coap_option_test.dart
```diff
@@ -87,86 +86,21 @@ void main() {
...
-
     test('Join', () {
-      final opt1 = CoapOption.createString(OptionType.uriPath, 'Hello');
-      final opt2 = CoapOption.createString(OptionType.uriPath, 'from');
-      final opt3 = CoapOption.createString(OptionType.uriPath, 'me');
-      final str = CoapOption.join(<CoapOption>[opt1, opt2, opt3], '/');
+      final opt1 = UriPathOption('Hello');
+      final opt2 = UriPathOption('from');
+      final opt3 = UriPathOption('me');
+      final str = Option.join(<Option<Object?>>[opt1, opt2, opt3], '/');
```
JosefWN, line 103: Tiny nitpick: I think we could omit the type for the list here and in other places, since it can be inferred?

```dart
final str = Option.join([opt1, opt2, opt3], '/');
```

Makes the code a little easier to read. Couldn't find a linter to enable for it though.

JKRhb: Removed the type annotation where possible :)

lib/src/option/integer_option.dart
JosefWN: A single line like the one you propose should work, but we need to use `Endian.big`, both when we encode and decode (`setUint16` and `getUint16`). That way we're saying it doesn't matter what `Endian.host` is, the byte data we work with is big/network endian.

The `ByteData` class will automatically reverse the list just like we did before, but only if `Endian.host == Endian.little`.

JosefWN: Maybe something like this would work?

```dart
ByteData.view(byteValue.buffer).getUint16(0, Endian.big);
```

JKRhb: Tried to incorporate this in e7b12fb51ed7f5d11e30a06ff687b2d231de95bf. The second argument of `getUint16` et al. defaults to `Endian.big`, so your latest comment should be addressed as well :) 

I added one comment/TODO due to the commit as I am not entirely sure about the case when you only have three bytes. In that case, you seem to need to extend the list of bytes before converting it to a 32bit (unsigned) integer. Do we need to take the endianness into account here as well?

JosefWN: The direction of the encoded data will always be the same regardless of architecture (it's only the direction of the decoded data that changes), so I think the padding should work. One thing that could be a little confusing is that the function would have side effects (`valueFromBytes` would mutate the caller's `byteValue`).

I think something like this should work, tried it out real quick outside of the project:
```dart
static int _valueFromBytes(final Uint8Buffer byteValue) {
  switch (byteValue.length) {
    case 0:
      return 0;
    case 1:
      return byteValue[0];
    case 2:
      return ByteData.view(byteValue.buffer).getUint16(0, Endian.big);
    case 3:
    case 4:
      final paddedBytes = Uint8List(4)..setAll(0, byteValue);
      return ByteData.view(paddedBytes.buffer).getUint32(0, Endian.big);
    default:
      final paddedBytes = Uint8List(8)..setAll(0, byteValue);
      return ByteData.view(paddedBytes.buffer).getUint64(0, Endian.big);
  }
}

static Uint8Buffer _bytesFromValue(final int value) {
  ByteData data;
  if (value < 0 || value >= (1 << 32)) {
    data = ByteData(8)..setUint64(0, value, Endian.big);
  } else if (value < (1 << 8)) {
    data = ByteData(1)..setUint8(0, value);
  } else if (value < (1 << 16)) {
    data = ByteData(2)..setUint16(0, value, Endian.big);
  } else {
    data = ByteData(4)..setUint32(0, value, Endian.big);
  }

  // RFC7252: A sender SHOULD represent the integer with as few bytes as possible,
  // i.e., without leading zero bytes (leading big endian being trailing).
  final buffer = Uint8Buffer()..addAll(data.buffer.asUint8List());
  while (buffer.isNotEmpty && buffer.last == 0) {
    buffer.removeLast();
  }

  return buffer;
}
```

JKRhb: Thanks @JosefWN! I incorporated your code in the latest commit :) However, there currently seems to be a bug regarding the trimming, as values like 256 become 1 at the moment.

JKRhb: Found the bug and will soon add an update :)

JKRhb: Bug has been fixed, the `trim()`  method could probably be refactored, though.

JKRhb: Applied some more refactoring, hopefully making the checks more readable.

JosefWN: There are a couple of reasons why I would use extension methods sparingly, they are mentioned in this article: https://daedtech.com/why-i-dont-like-c-extension-methods/

His main points I see in this context:
- _Decoupling class-specific functionality from that class (i.e. violating encapsulation)_, several points I see listed below are just side effects of that:
  - It makes it harder to deduct at first glance what is native to the `ByteData` class and what isn't (for example, I first looked at the class definition to see what `trim` does, and if it's trimming in both ends as opposed to `ltrim` or `rtrim`).
  - _The fact that you’re favoring free-floating static utility over instance methods._ In our case, the extension method would "leak" outside of this context whenever the file is imported, possibly causing confusion elsewhere.
  - _There is also the possibility for naming collisions if you write an extension method with the same signature as an instance method_ (for example if `ByteData` would introduce a `trim` method of its own).

I think they could be useful if you want to augment an API that you own (as opposed to the standard library), but perhaps not internally, as he argues later in the text:
> I’ve seen them used to ‘extend’ functionality in the same namespace where it would have been perfectly reasonable just to add a method to the class itself.

JKRhb: Reasonable points, I converted the extension methods to static methods and also got rid of an obsolete one.

JKRhb: In general, I've seen extension methods as a way to make code a bit more concise and readable (only using them for internal APIs), but I understand your concerns. Moreover, if they cause confusion about whether a method is defined within the respective class or not then my intended goal definitely has not been achieved here. Hope the updated version is better now :)

JKRhb:
Performed some more cleanup and resolved a few remaining issues. I think this PR should now be ready for another (final?) review :)

JKRhb:
Rebased the PR and provided a couple of additional test cases for invalid option formats.

JKRhb:
Added a last minute fix for the encoding of the OSCORE option.

JKRhb:
Rebased the PR against the latest upstream version. Now it should finally be ready to be merged :)

JKRhb:
Noticed an issue with the endianness of the integer option again... Should be fixed now. I am applying a couple of last changes, then this should actually be ready to be merged.

JKRhb:
I now also included a rework of the observe (de)registration values for requests in f9ea52f, resolving the last item from the checklist in #80.

PR #125, opened 2022-08-10
refactor: use HashMap for "parsing" numeric Content-Formats

JKRhb:
This PR refactors the creation of `CoapMediaType`s from numeric values by using a `HashMap` as a registry instead of iterating over all values. This should somewhat increase the performance, as `CoapMediaType`s can now be determined with constant instead of linear complexity.

### DIFF
--- a/lib/src/coap_media_type.dart
+++ b/lib/src/coap_media_type.dart
@@ -5,6 +5,7 @@
  * Copyright :  S.Hamblett
  */
 
+import 'dart:collection';
 import 'dart:io';
 import 'package:collection/collection.dart';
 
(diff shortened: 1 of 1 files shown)


lib/src/coap_media_type.dart
```diff
@@ -218,7 +219,7 @@ enum CoapMediaType {
   ),
 
   /// application/td+json
-  applicationTdJson(432, 'application', 'application/td+json'),
+  applicationTdJson(432, 'application', 'td+json'),
```
JKRhb, line 222: Noticed that there was a bug in the code which caused this Content-Format's string value to become `application/application/td+json`. Added a fix in 890ad1a7abe061109f98b19aa6b52585eee4ce49 and included it in this PR.

PR #127, opened 2022-08-13
refactor: clean up message decoding and encoding

JKRhb:
This PR makes a number of refactorings, getting rid of obsolete code for encoding and decoding, and simplifying the serialization/deserialization process. The simplification should pave the way for further cleanup and easier addition of new CoAP features, like the [extended token length RFC](https://www.rfc-editor.org/rfc/rfc8974) and the decoding/encoding of new message types like the signaling ones for CoAP over TCP.


lib/src/codec/message_decoder.dart
```diff
@@ -0,0 +1,226 @@
...
+    }
+    // Read options
+    var currentOption = 0;
+    while (reader.bytesAvailable) {
+      final nextByte = reader.readNextByte();
+      if (nextByte == specification.payloadMarker) {
+        if (!reader.bytesAvailable) {
+          // The presence of a marker followed by a zero-length payload
+          // must be processed as a message format error
+          throw const FormatException('Marker followed by 0 length payload');
```
JKRhb, line 84: In addition to the refactorings, I replaced the `StateError` here with a `FormatException` as otherwise, the error might crash an application if only exceptions are being caught.

JKRhb:
In contrast to #122, this PR could be merged right away. Besides the tests I also ran all examples, ensuring that they still work.

lib/src/coap_config.dart
```diff
@@ -27,16 +25,6 @@ abstract class DefaultCoapConfig {
   /// The version of the CoAP protocol.
   String get version => 'RFC7252';
 
-  /// The CoAP specification derived from the protocol version.
-  CoapISpec get spec {
```
JosefWN, line 31: I've only had a very quick glance at this PR, but I think we will need multiple encoders/decoders once we support TCP? Since the TCP wire format is different from UDP.

JKRhb: Oh, very good point, I didn't have such a close look into [relevant parts](https://datatracker.ietf.org/doc/html/rfc8323#section-3.2) of RFC 8323 yet. Having a specific encoder/decoder for each transport protocol definitely makes sense :)

JKRhb: I now tried to take into account that the existing Codec classes are UDP-specific in the latest commits.

lib/src/coap_config.dart
```diff
@@ -27,16 +25,6 @@ abstract class DefaultCoapConfig {
   /// The version of the CoAP protocol.
   String get version => 'RFC7252';
```
JosefWN, line 26: Are we using this now? If we assume that UDP is RFC7252 and TCP is RFC8323 we could infer the encoder/decoder to use from the URI. Not sure if the user would need to specify a `version` then?

lib/src/codec/udp/udp_message_decoder.dart
```diff
@@ -0,0 +1,225 @@
+/*
```
JosefWN, line 1: Perhaps we could shorten to `lib/src/codec/udp/message_decoder.dart`

lib/src/coap_config.dart
JKRhb: I think it is more of a leftover from the time when there were multiple draft versions the user could choose from. In practice, I think we can omit the version field, as I suppose that CoAP for UDP will stay at version 1 for the foreseeable future, while CoAP for TCP and WebSockets don't define the field in their message formats at all.

JKRhb: Initially, I already had it removed but then I was a bit unsure whether we should keep it. But I would also be in favor of dropping it :)

lib/src/codec/udp/udp_message_decoder.dart
```diff
@@ -0,0 +1,225 @@
...
+import '../../coap_message_type.dart';
+import '../../coap_option.dart';
+import '../../coap_option_type.dart';
+import '../../coap_request.dart';
+import '../../coap_response.dart';
+import 'datagram_reader.dart';
+import 'message_format.dart' as message_format;
+
+/// Provides methods to parse incoming byte arrays to messages.
+class CoapMessageDecoder {
```
JosefWN, line 23: We might need an interface for the encoder/decoder, if we have multiple encoders/decoders in the future it might make things a bit easier?

lib/src/codec/udp/udp_message_decoder.dart
JosefWN: Similarly for other files under `.../udp`, it's implicit by the folder they are in that they are dealing with UDP.

lib/src/coap_config.dart
JosefWN: I think it could be fine to remove it too, it could always be re-added later if multiple RFC's gain traction for the same network protocol

lib/src/codec/udp/udp_message_decoder.dart
JKRhb: I experimented a bit more with the encoder/decoder implementations and noticed that you can simplify the implementation quite drastically. I manage to boil down both serialization and deserialization to one class method each

Therefore we could also consider omitting the interfaces and define serialization/deserialization as plain functions. Alternatively, we could also turn them into (static) class methods or factory constructors on the `CoapMessage` class. Not sure what approach would be better here.

JKRhb: Tried out the function-based approach in the latest commit and makes things a bit easier I think. I also implemented the extended token length RFC while I was at it.

lib/src/codec/udp/message_encoder.dart
```diff
@@ -106,6 +117,61 @@ class UdpMessageEncoder implements MessageEncoder {
...
+  /// following [RFC 8974].
+  ///
+  /// [RFC 8974]: https://datatracker.ietf.org/doc/html/rfc8974
+  static int _getTokenLength(final Uint8Buffer? token) {
+    final tokenLength = token?.length ?? 0;
+    if (tokenLength <= 12) {
+      return tokenLength;
+    } else if (tokenLength <= 255 + 13) {
+      return 13;
+    } else if (tokenLength <= 65535 + 269) {
```
JosefWN: Not sure we can set the limit this high? Keep in mind that 65 535 is the MTU for an IP packet, the token length limit at 65535+269 would be larger than that.

This limit also includes IP headers which are added at the network layer. The _payload_ limit, which concerns us, I think is even lower:
> A UDP datagram is carried in a single IP packet and is hence limited to a maximum payload of 65,507 bytes for IPv4 and 65,527 bytes for IPv6. The transmission of large IP packets usually requires IP fragmentation.

https://erg.abdn.ac.uk/users/gorry/course/inet-pages/udp.html

JKRhb: Hmm, this should be the max value from [RFC 8974](https://datatracker.ietf.org/doc/html/rfc8974#section-2.1). However, there they state the following:

> The new definition of the "TKL" field increases the maximum token
> length that can be represented in a message to 65804 bytes.  However,
> the maximum token length that sender and recipient implementations
> support may be shorter.  For example, a constrained node of Class 1
> [[RFC7228](https://datatracker.ietf.org/doc/html/rfc7228)] might support extended token lengths only up to 32 bytes.
> 
> In CoAP over UDP, it is often beneficial to keep CoAP messages small
> enough to avoid IP fragmentation.  The maximum practical token length
> may therefore also be influenced by the Path MTU (PMTU).  See
> [Section 4.6 of [RFC7252]](https://datatracker.ietf.org/doc/html/rfc7252#section-4.6) for details.

Should we enforce a hard limit here? Or should we let users decide to handle this?

JosefWN: Huh, odd, let's go with the RFC then, I'm sure they have their reasons

lib/src/net/endpoint.dart
```diff
@@ -104,36 +102,30 @@ class Endpoint implements Outbox {
...
 
-      request!.source = event.address;
+    if (message == null) {
+      return;
+    }
+
+    if (message.needsRejection) {
+      _reject(message);
+      return;
+    }
```
JKRhb, line 115: @shamblett I have managed to find out that the ping issue is caused by this check. I will open a follow-up PR shortly.

PR #128, opened 2022-08-15
feat!: make client config optional

JKRhb:
In many scenarios, I got the impression that mandating the user to specify a custom config for each client is not really necessary. Instead, using the default config is actually fine and makes the library easier to handle if you need to create many requests to different endpoints at different places in your code.

Therefore, this PR proposes making the config parameter on the client optional, using the default config if the user did not supply one. This PR has the downside of creating a breaking API change, so it should probably only be merged before the release of version `6.0.0`.


shamblett:
This looks OK to me, I'll merge if you are happy.

JKRhb:
Awesome, feel free to merge :)

PR #129, opened 2022-08-16
feat!: introcuce `_sendWithStreamResponse` and `sendMulticast` methods, refactor send and observe implementations

JKRhb:
Thinking about a more elegant solution for performing multicast requests, I came up with introducing a special `sendStream` method on the client, which does not return a single `Future<CoapResponse>` but a `Stream` of responses library users can decide to listen to.

With this new API present, I realized that other aspects of the client could be refactored, too, making certain aspects of the client a bit more elegant. This includes the `CoapObserveClientRelation` class, which I refactored into an extension of the `Stream` class, increasing its usability.

All in all, I am quite happy with this new approach which should not only improve the usability for multicast scenarios but of the library as a whole.

### DIFF
--- a/example/get_observe_async.dart
+++ b/example/get_observe_async.dart
@@ -28,8 +28,8 @@ FutureOr<void> main() async {
   try {
     print('Observing /obs on ${uri.host}');
     final obs = await client.observe(reqObs);
-    obs.stream.listen((final e) {
-      print('/obs response: ${e.resp.payloadString}');
+    obs.listen((final e) {
+      print('/obs response: ${e.payloadString}');
     });
 
     final reqObsNon = CoapRequest(CoapCode.get, confirmable: false)
@@ -37,8 +37,8 @@ FutureOr<void> main() async {
 
     print('Observing /obs-non on ${uri.host}');
     final obsNon = await client.observe(reqObsNon);
-    obsNon.stream.listen((final e) {
-      print('/obs-non response: ${e.resp.payloadString}');
+    obsNon.listen((final e) {
+      print('/obs-non response: ${e.payloadString}');
     });
 
     final futures = <Future<void>>[];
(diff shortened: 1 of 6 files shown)


JosefWN:
Nice! `sendStream` sounds a little like we are _sending_ a stream of data, maybe we could have a private function called `_sendStreamResponse` (or something else which clarifies it's a streamed response), which we use for both multicast and observe, and then expose a `sendMulticast` method?

JKRhb:
> Nice! `sendStream` sounds a little like we are _sending_ a stream of data, maybe we could have a private function called `_sendStreamResponse` (or something else which clarifies it's a streamed response), which we use for both multicast and observe, and then expose a `sendMulticast` method?

That sounds very good, thank you! :) I'll incorporate it into the PR

JKRhb:
Updated the PR and slightly reworked the example to an actual multicast example with local CoAP nodes. I haven't tested it out yet, though. I now chose `_sendWithStreamResponse` as the name for the internal method to make it a bit clearer that the responses are returned as a Stream and not sent as one. Hope that it didn't become too verbose here.

shamblett:
Looks OK, the example still uses 'final conf = CoapConfig();' i.e not the default as updated in #128, no need to change this now, I'll do this when I merge this if #128 has been merged by then as is looking likely. Are you OK with this?

JKRhb:
Sure, thank you :) I could also rebase this one against the new `master` once #128 has been merged

JKRhb:
Rebased this PR against #128 for now, fixing a minor bug in the process, once #128 is merged I would rebase again, cleaning up the `fixup!` commit. Then we could also merge this one :)

shamblett:
#128 merged

JKRhb:
> #128 merged

Great, thank you :) Then this PR would also be ready to be merged now :)

Issue #121, opened 2022-08-03
Remove Coap prefix from internal classes


JosefWN:
More concise naming of internal classes would make the code a bit easier to read.

See: https://github.com/shamblett/coap/pull/113#discussion_r927217659

shamblett:
Believed complete with the release of version 6.0.0 of the package

PR #130, opened 2022-08-18
refactor: simplify interfaces and stack implementation

JKRhb:
This is yet another cleanup PR, getting rid of some more leftovers from the C# implementation. In particular, it removes some of the interface classes which are not actually needed and reworks the layers and stack implementation, making this part of the library more maintainable.

(This PR partly addresses #121.)


lib/src/coap_request.dart
```diff
@@ -86,13 +86,13 @@ class CoapRequest extends CoapMessage {
     _uri = value;
   }
 
-  CoapIEndPoint? _endpoint;
+  CoapEndPoint? _endpoint;
```
JosefWN, line 89: Could perhaps be renamed `CoapEndpoint` or even `Endpoint` (endpoint is one word but capitalized as two in `CoapEndPoint`).

lib/src/stack/layers/blockwise_layer.dart
```diff
@@ -10,26 +10,25 @@ import 'dart:math';
```
JosefWN, line 10: This file could perhaps be named just `blockwise.dart`, I would guess that it's a layer anyway since it's implied by the enclosing folder

JKRhb: Good point :) Should we remove the `layer` part from all files in the folder?

JKRhb: Renamed all the files in 92fa81cdc336ed21898e963f00c22178ccff3f0c for now

lib/src/coap_request.dart
JKRhb: Should be addressed now :) I renamed a couple of other files/classes as well. There are still some left which could be addressed in a follow-up PR, though.

PR #134, opened 2022-10-15
feat: add experimental CoAP server implementation

JKRhb:
I started working on a simple, experimental server implementation (for now only for UDP). For now, this is only a proof-of-concept, which can be improved upon once #127 has been merged.


JosefWN:
Cool!

Speaking of which, it would be really cool with integration tests against a server in the build pipeline/GitHub Actions. Re-writing our examples as integration tests and running them on push to a branch with an open PR... I was a bit hesitant to do this with coap.me or Californium since they are external and any issues with them would cause our pipelines to fail. Also having a third-party server like Coap.NET in our repo felt wrong.

It could be a lot of work to cover all examples, but even some would be a big win. Another option is to run them against Californium, which seems to be the best maintained, and allow for merging on failure if that's possible (or force merge on temporary issues).

Issue #132, opened 2022-09-03
Planning of next releases


JKRhb:
Considering the number of open PRs, I wanted to open this issue to discuss which PRs could go into which new release version. The open PRs in question are the following:

- [x] https://github.com/shamblett/coap/pull/122
- [x] https://github.com/shamblett/coap/pull/123
- [x] https://github.com/shamblett/coap/pull/124
- [x] https://github.com/shamblett/coap/pull/125
- [x] https://github.com/shamblett/coap/pull/127
- [x] https://github.com/shamblett/coap/pull/128
- [x] https://github.com/shamblett/coap/pull/129
- [x] https://github.com/shamblett/coap/pull/130
- [x] https://github.com/shamblett/coap/pull/133
- [x] https://github.com/shamblett/coap/pull/135
- [x] https://github.com/shamblett/coap/pull/136

I think https://github.com/shamblett/coap/pull/124 and https://github.com/shamblett/coap/pull/125 are trivial enough to actually become part of a 5.0.2 release, as they do not add a new feature or a breaking change. These could also be merged right away, I think.

https://github.com/shamblett/coap/pull/123 could become part of the next minor release, as it introduces new features without breaking changes.

The other PRs could become part of the next major release `6.0.0`, as they contain breaking changes and also apply quite significant changes to the project.

Do you think this roadmap makes sense, or do you think a different approach would be more reasonable? CC @JosefWN

shamblett:
Sounds like a plan to me, PR's 124 and 125 merged as suggested and package re published at version 5.0.2.

JosefWN:
Hey, sorry, been really busy. Sounds like a good plan.

JKRhb:
@shamblett Thank you for merging #123! As mentioned above, I think it could go into a version `5.1.0`.

Regarding the PRs that are still open, I think the easiest ones to review would be #128 and #129. Maybe you could have a look on them? As all remaining PRs contain breaking changes, I think we should consider them for a new major `6.0.0` release.

shamblett:
Package re published at version 5.1.0, I'll look at #128 and #129

JKRhb:
Thank you for merging these two! I think when it comes to the remaining PRs, it might make sense to review them in the following order (of increasing complexity): #130, #127, and #122. I just rebased them to the latest upstream version.

shamblett:
OK, #130 merged, #127 and #122 now have merge conflicts.

JKRhb:
[...]

PR #137, opened 2022-10-23
Fix timeout/max retransmission handling

JKRhb:
This PR provides a solution for the bug uncovered by @shamblett in https://github.com/shamblett/coap/issues/132#issuecomment-1288068964. The filtering logic is now moved inside a `StreamTransformer` object, which adds exceptions to the stream on timeout or cancellation. Similar to your suggestion in the issue, the relevant events are pre-filtered using a new `CoapCompletionEvent` interface that is "implemented" by the `CoapResponse`, `CoapRequestTimeoutException`, and `CoapRequestCancellationException` classes. Lastly, the PR applies minor refactoring to the matching of the original request and relevant responses.

### DIFF
--- a/lib/src/coap_client.dart
+++ b/lib/src/coap_client.dart
@@ -505,8 +505,8 @@ class CoapClient {
     await _prepare(request);
 
     final stream = _eventBus
-        .on<CoapRespondEvent>()
-        .map((final event) => event.resp)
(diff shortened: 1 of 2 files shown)


JKRhb:
Hmm. I just noticed that the cancellation event should actually not cause an exception to be added to the stream. I will adjust the PR a little.

Issue #140, opened 2022-10-29
Address TODO comments across the code base


shamblett:
There are a a few TODO comments building up across the code base, see below, these need to be either fixed or removed if no longer needed.

 
Found 17 TODO items in 6 files

coap_link_format.dart
(22, 6) // TODO(JKRhb): Revisit typing
(63, 6) // TODO(JKRhb): This needs to be a value between 1 and 4294967295.

integer_option.dart
(27, 8) // TODO(JKRhb): The handling of endianness should be revisited here.
(46, 8) // TODO(JKRhb): The handling of endianness should be revisited here.
(190, 4) // TODO(JKRhb): Is this really a class U option?
(228, 4) // TODO(JKRhb): Is this really a class E option?
(238, 4) // TODO(JKRhb): Is this really a class E option?

oscore_option.dart
(8, 4) // TODO(JKRhb): This currently can only contain encoded values and does not
(22, 6) // TODO(JKRhb): Maybe this can be done more elegantly

coap_client.dart
(40, 4) // FIXME: The name MatchETags might be a bit misleading, c.f. https://datatracker.ietf.org/doc/html/rfc7252#section-5.10.8.2

coap_media_type.dart
(336, 6) // TODO(JKRhb): Are there any uncovered cases?

coap_message.dart
(107, 10) // TODO(JKRhb): Revisit conditions for rejection
(238, 6) // TODO(JKRhb): Should rejections be included here as well?
(430, 6) // TODO(JKRhb): Apply proper percent-encoding
(469, 6) // TODO(JKRhb): Apply proper percent-encoding
(512, 6) // TODO(JKRhb): Apply proper percent-encoding
(560, 6) // TODO(JKRhb): Apply proper percent-encoding

PR #142, opened 2022-10-30
Address Todos

JKRhb:
This PR addresses some of the TODOs mentioned in #140 and a couple of minor linting issues.

As part of this PR, the `isPrintable` method is removed from the `CoapMediaType` enum as it did not fulfill a real purpose IMHO.

### DIFF
--- a/lib/src/coap_client.dart
+++ b/lib/src/coap_client.dart
@@ -36,13 +36,20 @@ import 'option/empty_option.dart';
 import 'option/integer_option.dart';
 import 'option/option.dart';
 
-/// The matching scheme to use for supplied ETags on PUT
-// FIXME: The name MatchETags might be a bit misleading, c.f. https://datatracker.ietf.org/doc/html/rfc7252#section-5.10.8.2
+/// The strategy applied for conditional requsts.
+///
+/// Can either specifiy that one of the supplied ETags must match the
+/// target resource [onMatch] or that the target resource must not exist
+/// [onNoneMatch].
+///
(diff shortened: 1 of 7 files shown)


JKRhb:
Covered two more TODOs with the latest commits.

lib/src/network/coap_network_tinydtls.dart
```diff
@@ -77,7 +77,7 @@ class CoapNetworkUDPTinyDtls extends CoapNetworkUDP {
     super.port,
     super.bindAddress,
     this._tinyDtlsInstance, {
-    super.namespace,
+    final super.namespace,
```
JosefWN, line 80: I think when I upgraded to Flutter 3.3 (Dart 2.18) or possibly VS Code's Flutter/Dart plugin, a linter rule was added to exclude the need for final super. Maybe we should add it to our lint config too: https://dart.dev/tools/diagnostic-messages?utm_source=dartdev&utm_medium=redir&utm_id=diagcode&utm_content=unnecessary_final#unnecessary_final

JKRhb: Thank you! :) I actually haven't had the latest Dart version installed :)

JKRhb:
@shamblett Did you maybe have the chance to have a look on this PR and https://github.com/shamblett/coap/pull/141 yet?

shamblett:
OK, they look OK to me, I'll merge them if you agree.

JKRhb:
> OK, they look OK to me, I'll merge them if you agree.

Sure, that would be great :)

Issue #147, opened 2022-12-23
openssl (dtls) messages are not processed 


vincent-iQontrol:
When we try to read a value like:  await client.put('Json', payload: payload); messages are comming in network/coap_network_openssl.dart in the _dtlsConnection?.received.listen part, but only if I add a delay before the listener like this:

`await Future.delayed(Duration(milliseconds: 2000));
    _dtlsConnection?.received.listen(
      (final frame) {
        final message =
            CoapMessage.fromUdpPayload(Uint8Buffer()..addAll(frame));
        eventBus.fire(CoapMessageReceivedEvent(message, address));`

The problem is that it looks like the messages are not pushed to the eventBus because there is nothing coming in on the net/endpoint.dart `void _receiveMessage(final CoapMessageReceivedEvent event) {`

Resulting in a `CoAP encountered an exception: CoapRequestTimeoutException: Request timed out after 2 retransmits.
` message

vincent-iQontrol:
I found the issue with the eventbus, namespace wasn't passed into coap_inetwork.dart:

```
case DtlsBackend.OpenSsl:
            return CoapNetworkUDPOpenSSL(
              address,
              port ?? config.defaultSecurePort,
              bindAddress ?? defaultBindAddress,
              verify: config.dtlsVerify,
              withTrustedRoots: config.dtlsWithTrustedRoots,
              ciphers: config.dtlsCiphers,
              namespace: namespace,
              rootCertificates: config.rootCertificates,
            );
```

JKRhb:
Oh, very good catch! Do you want to create a PR for fixing the issue? :) Or should I update #146?

vincent-iQontrol:
> 

I did a PR

JKRhb:
@vincent-iQontrol Thank you once again for resolving the issue :) Did you also require changes to the `dtls` package by the way? Or does it work with the version currently available from pub.dev?

JKRhb:
@vincent-iQontrol FYI: In OpenSSL 3.2, a new BIO method will be introduced, which should resolve the issues with the fragmented handshake messages (which will make it unnecessary to limit the number of ciphers). That should probably make things a lot easier :)

vincent-iQontrol:
> @vincent-iQontrol Thank you once again for resolving the issue :) Did you also require changes to the `dtls` package by the way? Or does it work with the version currently available from pub.dev?

No changes necessary in dtls package at the moment

vincent-iQontrol:
[...]

PR #148, opened 2022-12-23
openssl (dtls) messages are not processed #147

vincent-iQontrol:
- fixed missing namespace
- moved DtlsClientConnection before binding to fix listen issues


JKRhb (approved):
@shamblett Could you have a look on this PR and #146? They should be mergeable right away :)

shamblett:
OK done, do you want a release for this? If so I'll do this tomorrow.

JKRhb:
I think a new release would be nice :) However, I guess it would be another major one due to some API changes, right?

shamblett:
Yep we need a new one, 4 other PR's have gone into this also, version 7.0.0 published, a COAP package for xmas!

PR #149, opened 2022-12-28
feat!: replace `dtls` with `dtls2`, remove `dart_tinydtls`

JKRhb:
Inspired by @vincent-iQontrol, I created a fork of the discontinued `dtls` package called [`dtls2`](https://pub.dev/packages/dtls2) and added PSK functionality to it. The implementation is still pretty much experimental, but it might be a good testing ground for future progress in this area. I would be grateful if you could test it out and provide feedback if you notice that something's not working or in need of improvement :)


JKRhb:
I now applied some additional refactorings/reworks in my dtls2 fork, making the implementation a bit more solid. It would be great if you could have a look/test the PR out :)

JKRhb:
Besides Linux, the OpenSSL network should now also support macOS and Windows. Could you maybe try it out, @JosefWN and @shamblett? :)

JosefWN:
One gripe:
```yaml
dependencies:
  ffi: ^1.1.0
```

Depending on the package makes it trickier to use other packages which have long since moved to `ffi: ^2.0.0`.

JKRhb:
I am actually updating the package's at the moment :) However, there have been some issues with dart_tinydtls and ffi 2.x.x, so the two packages might not be compatible anymore :/ Maybe we can also drop support for dart_tinydtls, though, since dtls2 already covers most of the client-related features (except for Raw Public Key support).

JKRhb:
After debugging dart_tinydtls for a long time due to strange bugs that only occurred on Windows when trying to update the ffi package to version 2.0.1, I now decided to make a hard cut and simply remove it as a dependency. It's a bit unfortunate that this also removes RPK support from the package, but hopefully the feature gets added to OpenSSL in the upcoming months (there is currently a PR that is WIP). If I should find the time to properly fix dart_tinydtls, maybe we can also add it back in. However, using OpenSSL is of course much more appropriate for non-constrained devices.

Another drawback is also that dtls2 does not have a DTLS server implementation yet – but I am working on it :)

lib/src/coap_config.dart
```diff
@@ -98,4 +83,12 @@ abstract class DefaultCoapConfig {
...
   /// Can be used to specify the Ciphers that should be used by OpenSSL.
   String? get dtlsCiphers => null;
+
+  /// Custom libssl instance that can be registered if OpenSSL
+  /// should not be available at the default locations.
+  OpenSsl? get libSslInstance => null;
+
+  /// Custom libcrypto instance that can be registered if OpenSSL
+  /// should not be available at the default locations.
+  OpenSsl? get libCryptoInstance => null;
```
JKRhb, line 93: I am wondering if the type of these should rather be changed to `DynamicLibrary`. This would also get rid of the `OpenSsl` reexport.

JKRhb: Changed this now in the latest version of dtls2.

JKRhb:
After a number of additional iterations and dtls2 versions, I think this PR should now be ready for a final review :)

JKRhb:
Just updated `dtls2` to version 0.9.0, which now also features DTLS server functionality :)

JKRhb:
Applied some additional cleanup and updated `dtls2` to version 0.10.0.

vincent-iQontrol:
@JKRhb replace? Does that mean openssl with certificates will disappear?

JKRhb:
> @JKRhb replace? Does that mean openssl with certificates will disappear?

All of the current OpenSSL features should actually remain (with added PSK support), only the Raw Public Key support provided by tinydtls should be removed. Did certificates stopped working with the changes present in this PR? If so, then it wasn't intentional and will be fixed :)

vincent-iQontrol:
I did test it, it work. I did have to override the collection package because flutter uses this version. I will report if I find something. 

dependency_overrides:
   collection: ^1.16.0

JKRhb:
@shamblett Do you maybe have time in the upcoming days to review some of the recent PRs? :)

shamblett:
Yep I'll have a look tomorrow

JKRhb:
(Rebased to the latest upstream master branch version.)

JKRhb:
@shamblett I think this PR should now be ready to be merged as well. The same goes for (the rather trivial) #161.

PR #157, opened 2023-01-19
fix: apply proper percent encoding to URI options

JKRhb:
This PR applies some fixes to the URI options (regarding percent-encoding) and resolves a problem with the String option encoding.

It also resolves a couple of TODOs from #142.

PR #158, opened 2023-01-21
feat: only skip host and port options if non-default

JKRhb:
Reading through the RFC, I noticed that the skipping of the Uri-Host and Uri-Port options is not necessary if their values are not the default. Including them as proposed in this PR makes it easier to determine where a request was sent to, as illustrated in the following Wireshark screenshot:

grafik

(This PR depends on #157 and should only be merged afterwards.)


shamblett:
If your happy this is complete I'll merge it

PR #160, opened 2023-01-22
feat!: set URI for each request instead of client

JKRhb:
I started a new attempt at moving the URL argument from the client to the request methods, making it easier to reuse clients for multiple requests. In theory, this should also make it more efficient to make multiple requests to different endpoints, as they can all use the same socket(s).

At the moment, this PR is still WIP and should be considered experimental, since for now only UDP is covered. It will also require more discussion and needs to address the concerns raised in #93. I will try to come up with a proper solution after #149 is merged, as the `dtls2` API should make it much easier to also implement this PR's approach for DTLS.

### DIFF
--- a/example/cancel_request.dart
+++ b/example/cancel_request.dart
@@ -16,14 +16,14 @@ import 'config/coap_config.dart';
 FutureOr<void> main() async {
   final conf = CoapConfig();
   final uri = Uri(scheme: 'coap', host: 'coap.me', port: conf.defaultPort);
-  final client = CoapClient(uri, config: conf);
+  final client = CoapClient(config: conf);
 
-  final cancelThisReq = CoapRequest.newGet()..uriPath = 'doesNotExist';
+  final cancelThisReq = CoapRequest.newGet(uri.replace(path: 'doesNotExist'));
 
   try {
     // Ensure this request is not also cancelled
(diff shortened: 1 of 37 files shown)


JKRhb:
I think this PR should now be ready for an initial review/testing :)

JKRhb:
I think this PR should now be finished for the time being and ready for thorough review. In a follow-up, the TCP network should be fixed/actually completed (I already prepared something locally). Furthermore, the API needs to be adjusted for potentially reconnecting to a peer (for example, if session reuse is discouraged as in the context of ACE-OAuth, c.f. [Section 7.1 of RFC 9202](https://datatracker.ietf.org/doc/html/rfc9202#section-7.1)).

JKRhb:
Converted this PR back to a draft since I noticed an issue with the reliability layer in the case of CoAP over DTLS.

JKRhb:
> Converted this PR back to a draft since I noticed an issue with the reliability layer in the case of CoAP over DTLS.

Found a way to fix this and added some additional cleanup.

JKRhb:
Before considering this PR for merging, I guess it would be better to first merge #169 and then spin off some of the unrelated changes into smaller PRs. However, this PR can already work as a proof of concept that setting the URL for each request works.

PR #164, opened 2023-01-25
refactor: refactor option list printing

JKRhb:
Dealing with some more potential refactoring, this PR makes a proposal for reworking the printing of the options contained within the different message types.


JKRhb:
Found a way to generalize the printing of the options and to clean up the code a bit more. Should now be ready to be merged.

PR #168, opened 2023-01-28
refactor: replace OSCORE Interfaces with Mixins

JKRhb:
This PR proposes a trivial refactoring, turning the interfaces for identifying OSCORE class into more idiomatic Mixins.

### DIFF
--- a/lib/src/option/coap_block_option.dart
+++ b/lib/src/option/coap_block_option.dart
@@ -85,7 +85,7 @@ enum BlockOptionType {
 
 /// This class describes the block options of the CoAP messages
 abstract class CoapBlockOption extends IntegerOption
-    implements OscoreOptionClassE, OscoreOptionClassU {
+    with OscoreOptionClassE, OscoreOptionClassU {
   String get _szxErrorMessage =>
       'Encountered reserved SZX value 7 in $runtimeType.';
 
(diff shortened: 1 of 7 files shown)


JKRhb:
@shamblett I think this PR as well as #162, #163, #164, #165, and #167 are rather trivial and could be merged right away.

With #160, I made a lot of progress, there are some minor aspects that still need fixing. There might also be some discussion required if this is the general direction we should go for. However, I think being able to have only client for requests to multiple endpoints should be handy, especially once we also start to support TCP and WebSockets. Being able to use only one socket for multiple peers should also increase the efficiency of the implementation if you have a lot of devices you want to communicate with.

shamblett:
All done except for #167 which has a conflict, also there have been 11 PR merges since 7.0.0 so I'm thinking we should have another major release at this point.

JKRhb:
> All done except for #167 which has a conflict, also there have been 11 PR merges since 7.0.0 so I'm thinking we should have another major release at this point.

Awesome, thank you! :) #167 should now be ready as well. A new major release sounds very good! For 9.0.0 we can then concentrate on the discussion of new major features (TCP, WebSockets, maybe even OSCORE).

shamblett:
All done, version 8.0.0 published.

PR #169, opened 2023-01-31
feat!: rework URI handling for CoAP requests and responses

JKRhb:
This PR proposes a rework of the `uri` field within the `CoapRequest` class. Before, the `Uri` was constructed by using individual `Uri-Host`, `Uri-Path`, `Uri-Port`, and `Uri-Query` options.

However, I noticed that a request's URI can be set in a much simpler way by simply turning it into a regular field which can then be converted into and from individual URI component options. This enabled me to implement the algorithms described in [section 6.4](https://www.rfc-editor.org/rfc/rfc7252.html#section-6.4) and [section 6.5 of RFC 7252](https://www.rfc-editor.org/rfc/rfc7252.html#section-6.5) in two functions, which makes debugging a bit easier. With these changes, the conversion should now be RFC-compliant (again) and slightly easier to maintain.

The new approach also allowed/incentivized me to make a slight (but breaking) API change, adding the URI as a request constructor parameter. Therefore, this PR can also be seen as an intermediate step towards #160, but also towards further API and type-safety improvements.

(CC @shamblett @JosefWN)

### DIFF
--- a/example/cancel_request.dart
+++ b/example/cancel_request.dart
@@ -18,7 +18,7 @@ FutureOr<void> main() async {
   final uri = Uri(scheme: 'coap', host: 'coap.me', port: conf.defaultPort);
   final client = CoapClient(uri, config: conf);
 
-  final cancelThisReq = CoapRequest.newGet()..uriPath = 'doesNotExist';
+  final cancelThisReq = CoapRequest.newGet(uri.replace(path: 'doesNotExist'));
 
   try {
     // Ensure this request is not also cancelled
--- a/example/get_max_retransmit.dart
+++ b/example/get_max_retransmit.dart
@@ -20,7 +20,7 @@ FutureOr<void> main() async {
 
   print('maxRetransmit config: ${conf.maxRetransmit}');
 
-  final request = CoapRequest.newGet()..uriPath = 'doesNotExist';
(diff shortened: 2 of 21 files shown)


JKRhb:
I also added a rework for the location options of the `CoapResponse` class in the latest commits, reusing code from the URI options. This part of the implementation should now also be a bit more RFC-compliant.

JKRhb:
There are still some details from RFC 7252 that should be addressed later, but I think for now this PR should be ready for review/merging.

JKRhb:
@shamblett Did you have the chance to look at this PR yet? I think merging it would pave the way for a lot of further improvements. #170, #172, and #174 would also be ready to be merged but might be a bit simpler to review than this one.

shamblett:
Yep OK merged, the discover_resources and get_blockwise examples seem to be hanging now when I run them, this may be my machine here, I'll have a closer look tomorrow.

JKRhb:
> Yep OK merged, the discover_resources and get_blockwise examples seem to be hanging now when I run them, this may be my machine here, I'll have a closer look tomorrow.

Thank you! But you are right, there seems to be a blockwise issue which I haven't noticed :( I try to come up with a fix as soon as possible :)

JKRhb:
Found the cause of the issue and opened #175 as a quick fix.

PR #171, opened 2023-02-02
chore: let CI generate code coverage reports

JKRhb:
This PR adjusts the CI pipeline slightly so that it generates code coverage reports when testing. This would make it possible to visualize coverage of the code base using a provider such as codecov or coveralls in a next step.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,5 +26,5 @@ jobs:
       - name: Analyze project source
         run: dart analyze
 
-      - name: Run tests
-        run: dart test
+      - name: Run tests with coverage
+        run: dart run coverage:test_with_coverage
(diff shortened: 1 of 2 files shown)


JKRhb:
@shamblett Thank you merging! I think #170 and #172 should also be easy to review. #169 is a bit more complex, but should also simplify things and reduce the API surface.

JKRhb:
Also: Would adding something like codecov be an option for you? Using it would make it a bit easier to see which parts of the library are covered by tests.

shamblett:
Yes no problem with this. I did use to include code coverage of my packages years ago but the tooling for it in Dart changed a lot so I removed it, seems to be better again now so there's no reason why it can't be re-instated.

PR #173, opened 2023-02-06
chore(ci): add codecov upload step

JKRhb:
As discussed in #171, this PR adds an upload step for codecov.io which enables code coverage visualization features. ~~For this to properly work, adding a `CODECOV_TOKEN` as a secret would also be needed, I think.~~

Using the [codecov GitHub app](https://github.com/marketplace/codecov), we could integrate code coverage report into PRs, which I quite like since they create an incentive to provide additional tests. Not everyone seems to be a fan of them, though.

Edit: Seems as if the bot is working already :)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,3 +28,11 @@ jobs:
 
       - name: Run tests with coverage
         run: dart run coverage:test_with_coverage
+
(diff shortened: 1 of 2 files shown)


JKRhb:
While I was at it, I also included two more badged in the README file.

PR #175, opened 2023-02-16
fix: fix blockwise transfer

JKRhb:
#169 introduced a small bug that broke blockwise transfer, since request URIs were copied twice. This PR provides a quick fix that should be followed up by a more elegant solution in the future, though.


shamblett:
Yep, all working now thanks. I'll start looking at the the other PR's over the weekend.

PR #176, opened 2023-02-16
Improve request APIs

JKRhb:
This PR makes a couple of improvements to the request API by adding a `query` parameter, adding payload-related parameters to the `fetch` method (which actually is the main USP of this request method), and changing the type of the payload parameter from `Uint8Buffer` to `Iterable<int>`, increasing the number of possible input types (by including types like `Uint8List` or `List<int>`).

The PR also applies some additional clean up, getting rid of some getters and setters that are not actually needed.

### DIFF
--- a/lib/src/coap_client.dart
+++ b/lib/src/coap_client.dart
@@ -183,7 +183,9 @@ class CoapClient {
       ),
       confirmable: confirmable,
       accept: accept,
-    )..setPayloadMedia(payload, format);
+      contentFormat: format,
+      payload: utf8.encode(payload),
+    );
     _build(
       request,
       options,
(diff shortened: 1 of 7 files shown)


JKRhb:
Note: In order to avoid a merge conflict, this PR depends on #175 and should be merged afterward.

JKRhb:
@shamblett Did you have the chance to have a look at this PR yet?

shamblett:
I'll have a look at this over the weekend.

shamblett:
Looks good, merged. There's about 9 PR's merged now since version 8.0.0, may be an appropriate point for the next release.

JKRhb:
Great, thank you, @shamblett! :) I have two smaller PRs inflight which I would open shortly. Afterward, a new release sounds very good :)

PR #177, opened 2023-02-18
WIP: Implement CoAP over TCP 

JKRhb:
As mentioned in some other PR, I started working on improving the CoAP over TCP implementation, adding the necessary logic for exchanging messages. Connecting to a (plain) TCP server and sending and receiving messages already works.

However, responses cannot be matched to requests yet, since the matching logic has to be slightly adjusted (CoAP over TCP messages don't contain message ID or type fields). Furthermore, the layer stack needs to take into account that resending messages isn't needed for reliable transport. The implementation is also lacking support for signaling messages, which need to be sent at the beginning of an exchange. Lastly, I currently perform three stream transformations in a row for incoming TCP segments, which could probably improved efficiency-wise.

So there is still plenty to do, which I hope to deal with over the next couple of weeks.


JKRhb:
I now actually managed to finish the basic logic for connecting and communicating with a server. There is still some more polish required, though.

BastienGermond:
Hi @JKRhb, thank you a lot for adding support for CoAP over TCP. Is it something that you plan to keep working on ? I was wondering if the current state does fully support TCP ?

JKRhb:
Hi @BastienGermond! I definitely plan to continue working on this :) However, it might only be in a couple of weeks :/ The current state is not really working yet IIRC, but I think you are able to at least send and receive messages – I'll try to have another look at it as soon as possible :)

PR #182, opened 2023-06-23
chore: update dtls2 to version 0.15.0

JKRhb:
I just released a new version of `dtls2` with a few fixes. This PR incorporates updates the dependency here.


shamblett:
Looks Ok, merging

PR #183, opened 2023-07-11
chore: upgrade `dtls2` to version 0.16.0

JKRhb:
I just released a new version of the `dtls2` package with some additional bug fixes. This PR bumps the version number.


shamblett (approved):
OK thanks.

PR #189, opened 2024-05-22
Update dtls2 package to version 0.17.0

JKRhb:
This PR updates the dtls2 package to the latest version.

There are currently a couple of issues when trying to use the CoAPS example with newer versions of DTLS because they have deprecated a number of cipher suites, including the mandatory one for CoAP (`TLS_PSK_WITH_AES_128_CCM_8`), making it necessary to explicitly state that you want to them, which is a bit annoying. I will create another patch version soon that will address this issue and make the package usable in this context again.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -14,7 +14,7 @@ dependencies:
   build: ^2.4.1
   collection: ^1.18.0
   convert: ^3.1.1
-  dtls2: ^0.16.0
+  dtls2: ^0.17.0
   event_bus: ^2.0.0
   meta: ^1.12.0
   path: ^1.9.0


shamblett:
OK thanks ,merged, as you can see there is a build error on Windows, all the tests are passing so I'm guessing the failure(which looks a bit weird) may have nothing to do with this PR, happy for this to be addressed in your future patch.

JKRhb:
Thank you for merging this PR! :)

The CI issue on Windows is an upstream problem that exists since the release of Dart 3.4 (see https://github.com/dart-lang/sdk/issues/55745). The workaround for now seems to be switching back to Dart 3.3 on Windows until the issue has been fixed. But since the tests are passing, this is probably not strictly necessary and will hopefully be resolved soon.

PR #191, opened 2024-05-27
ci: add workaround for failing CI on Windows

JKRhb:
As mentioned in #189, there is currently an issue with the CI when it is running under Dart 3.4.0 and Windows. This issue adds a workaround that lets the CI fall back to 3.3.0 under Windows so that the respective job does not fail anymore due to a timeout. As soon as the issue has been fixed upstream, this change can be reverted.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,11 +11,24 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
+        # TODO: Revert once https://github.com/dart-lang/sdk/issues/55745
+        #       has been resolved
+
+        sdk: [stable, 3.3.0]
+        exclude:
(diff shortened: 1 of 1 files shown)


JKRhb:
Hmmm. For some reason, lowering the version to 3.3.0 on Windows does not help here 🤔 It's a bit strange since in another project, that actually made the CI pass again. I will investigate why that might be the case.

shamblett:
Yep thanks, I've seen weird behaviour like this before over the years, usually more in browser tests, sometimes it just seems to go away, also in one instance upgrading to the latest version of the test package cleared the hang.

JKRhb:
Sorry, I realized only now that there was an essential line missing for this workaround to actually work 🤦 Now this PR should be ready for merging :)

shamblett:
Thanks, package re released at version 9.0.5

Issue #193, opened 2024-06-07
First response from observe not being handled by the onData function passed as argument


jorgesanfz:
In the observe method, the check for the observe flag within the `CoapResponse` involves awaiting `_waitForResponse`. This delays the return of the `CoapObserveClientRelation` object until after the first response is received. As a result, the `CoapObserveClientRelation` object does not yet have the `onData` function passed as an argument, and the first response cannot be handled.

shamblett:
Thanks, looks OK merged.

shamblett:
Package re released at version 9.0.5

PR #195, opened 2024-06-07
Fix: first observe response not being handled

jorgesanfz:
PR related to issue #193

To prevent the check from blocking the function until after the first response is received, I created a method within the `CoapObserveClientRelation` called `checkObserve` that checks for the observe flag upon receiving the first response. This approach eliminates the need for await, ensuring that the `CoapObserveClientRelation` object already has our `onData` method when the first response is received.


shamblett:
OK thanks, seems to work OK, merged, will be in the next release.

Note don't worry about the PR failing on windows, this is known and being worked on.

PR #200, opened 2024-12-11
chore!: update dtls2 package to version 0.18.0

JKRhb:
This PR updates the `dtls2` package to the latest version. Since there have been API changes in `dtls2` regarding the handling of certificates, the relevant parts of the API for this package are modified accordingly and the certificate classes in question are re-exported.

The main benefit of this change is now that you are able to provide root certificates that can also be in PEM format, making the use of DTLS in PKI mode a bit more versatile and the API itself clearer. However, as there are API changes involved, they have to be considered breaking changes.

### DIFF
--- a/lib/coap.dart
+++ b/lib/coap.dart
@@ -7,6 +7,9 @@
 
 library coap;
 
+export 'package:dtls2/dtls2.dart'
+    show Certificate, PemCertificate, DerCertificate;
+
 /// Pre-defined configs
 export 'config/coap_config_all.dart';
 export 'config/coap_config_default.dart';
--- a/lib/src/coap_config.dart
+++ b/lib/src/coap_config.dart
@@ -7,6 +7,7 @@
 
 import 'dart:typed_data';
 import 'dart:ffi';
+import 'package:dtls2/dtls2.dart';
 
 import 'coap_constants.dart';
 import 'deduplication/deduplicator_factory.dart';
@@ -79,7 +80,7 @@ abstract class DefaultCoapConfig {
   bool get dtlsWithTrustedRoots => true;
 
   /// List of custom root certificates to use with OpenSSL.
-  List<Uint8List> get rootCertificates => const [];
+  List<Certificate> get rootCertificates => const [];
 
   /// Can be used to specify the Ciphers that should be used by OpenSSL.
   String? get dtlsCiphers => null;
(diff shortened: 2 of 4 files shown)


shamblett:
Getting some analysis errors on merge -

error: The name 'Certificate' isn't a type, so it can't be used as a type argument. (non_type_as_type_argument at [coap] lib/src/network/coap_network_openssl.dart:61)
error: The name 'Certificate' isn't a type, so it can't be used as a type argument. (non_type_as_type_argument at [coap] lib/src/network/coap_network_openssl.dart:81)
warning: The library 'package:dtls2/dtls2.dart' doesn't export a member with the shown name 'Certificate'. (undefined_shown_name at [coap] lib/coap.dart:11)
warning: The library 'package:dtls2/dtls2.dart' doesn't export a member with the shown name 'PemCertificate'. (undefined_shown_name at [coap] lib/coap.dart:11)
warning: The library 'package:dtls2/dtls2.dart' doesn't export a member with the shown name 'DerCertificate'. (undefined_shown_name at [coap] lib/coap.dart:11)

JKRhb:
Thank you for merging and your comment! Are you getting these errors locally or in the CI?

shamblett:
Hi,

Locally in my IDE(Intellij) appeared as soon as I merged into master.

JKRhb:
> Hi,
> 
> Locally in my IDE(Intellij) appeared as soon as I merged into master.

Hmm, that is strange, I cannot reproduce it on my machine, and the CI seems also fine – do you maybe need to rerun `dart pub get` and/or delete the `pubspec.lock` file?

shamblett:
Ah yes, needed a pub get! I wish it would remind me of this when I pull an updated pubspec.yaml.

shamblett:
Running example get_resource_secure.dart gives -
```
Sending get /test to californium.eclipseprojects.io
CoAP encountered an exception: DtlsException: No ciphers available. If you are using PSK cipher suites, check you have defined a pskCredentialsCallback.
```

Probably just needs an update for the DTLS changes.

JKRhb:
Thank you for pointing that out!  I think it is actually related to recent changes in OpenSSL which cause the cipher suite that is defined in the example to only be usable when reducing the "security level". I've already implemented a way to adjust the security level in dtls2, for which I will create a PR here shortly.

shamblett:
Great thanks.

JKRhb:
#203 should now provide a fix for this issue :)

PR #203, opened 2024-12-12
Allow for overriding the OpenSSL security level

JKRhb:
As discussed in https://github.com/shamblett/coap/pull/200, this PR adds the ability to override the security level used by OpenSSL.  This allows for still using cipher suites like `TLS_PSK_WITH_AES_128_CCM_8` that must be supported by CoAP implementations (see [section 9.1.3.1](https://datatracker.ietf.org/doc/html/rfc7252#section-9.1.3.1) of RFC 7252), as more recent versions of OpenSSL are not allowing them to be used anymore without lowering the security level.

The DTLS example is adjusted accordingly so that it now works again.

### DIFF
--- a/example/get_resource_secure.dart
+++ b/example/get_resource_secure.dart
@@ -30,6 +30,14 @@ PskCredentials pskCredentialsCallback(final String? identityHint) =>
 class DtlsConfig extends DefaultCoapConfig {
   @override
   String? get dtlsCiphers => 'PSK-AES128-CCM8';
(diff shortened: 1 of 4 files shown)


shamblett:
Great thanks, all working now. Package re released at version 9.1.0.

JKRhb:
> Great thanks, all working now. Package re released at version 9.1.0.

Awesome, thank you! :)

